PR #56, opened 2015-01-02
allow datetime.timedelta for leeway arg

### LINKED ISSUE #54: allow datetime.timedelta instance for leeway argument
the summary says it all.

wbolster:
This fixes #54.

### DIFF
--- a/jwt/__init__.py
+++ b/jwt/__init__.py
@@ -11,7 +11,7 @@ import hashlib
 import hmac
 import sys
 
-from datetime import datetime
+from datetime import datetime, timedelta
 from calendar import timegm
 from collections import Mapping
 
@@ -359,6 +359,10 @@ def load(jwt):
 
 def verify_signature(payload, signing_input, header, signature, key='',
                      verify_expiration=True, leeway=0, **kwargs):
+
+    if isinstance(leeway, timedelta):
+        leeway = leeway.total_seconds()
+
     try:
         algorithm = header['alg'].upper()
         key = prepare_key_methods[algorithm](key)
--- a/jwt/__init__.py
+++ b/jwt/__init__.py
@@ -361,7 +361,14 @@ def verify_signature(payload, signing_input, header, signature, key='',
                      verify_expiration=True, leeway=0, **kwargs):
 
     if isinstance(leeway, timedelta):
-        leeway = leeway.total_seconds()
+        try:
+            leeway.total_seconds
+        except AttributeError:
+            # On Python 2.6, timedelta instances do not have
+            # a .total_seconds() method.
+            leeway = leeway.days * 24 * 60 * 60 + leeway.seconds
+        else:
+            leeway = leeway.total_seconds()
 
     try:
         algorithm = header['alg'].upper()
--- a/README.md
+++ b/README.md
@@ -151,6 +151,13 @@ time.sleep(32)
 jwt.decode(jwt_payload, 'secret', leeway=10)
 ```
 
+Instead of specifying the leeway as a number of seconds, a `datetime.timedelta` instance can be used. The last line in the example above is equivalent to:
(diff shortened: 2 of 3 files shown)


wbolster:
Btw, the tests directly call into `verify_signature()`, while this doesn't seem to be part of the public API (not in `__all__`). Is this intentional?

wbolster:
Sigh, Python 2.6 doesn't have `datetime.timedelta.total_seconds()`. I'll add a compatibility hack.

jpadilla:
@wbolster LGTM. Let's also update the README to document this behavior.

wbolster:
ok will update the readme with examples

"José Padilla" <notifications@github.com> schreef op 2 januari 2015 19:59:36 CET:
>@wbolster LGTM. Let's also update the README to document this behavior.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/progrium/pyjwt/pull/56#issuecomment-68551872

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

jwt/__init__.py
```diff
@@ -7,11 +7,11 @@
 from __future__ import unicode_literals
 import base64
 import binascii
+import datetime
 import hashlib
 import hmac
 import sys
 
-from datetime import datetime
```
mark-adams: I think this would be cleaner if we kept this line and added timedelta

jwt/__init__.py
```diff
@@ -359,6 +359,17 @@ def load(jwt):
 
 def verify_signature(payload, signing_input, header, signature, key='',
                      verify_expiration=True, leeway=0, **kwargs):
+
+    if isinstance(leeway, datetime.timedelta):
+        try:
+            leeway.total_seconds
```
mark-adams: What about replacing the try-catch with 
```
if hasattr(leeway, 'total_seconds') and callable(leeway.total_seconds):
    leeway = leeway.total_seconds()
else:
    leeway = leeway.days * 24 * 60 * 60 + leeway.seconds
```

I think this would avoid overhead caused by the exception handling and be a little clearer to the reader.

jwt/__init__.py
mark-adams: Specifically, it doesn't make sense to add `datetime.` in 10+ places when we can just add timedelta to this import.

jpadilla: Agree with this.

jwt/__init__.py
jpadilla: Not sure about this one though, I usually prefer to handle things like this via exceptions.

wbolster: At least in CPython, exceptions are only "expensive" when they actually occur, which would only be on Python 2.6 in this case, so I agree with @jpadilla that my approach using an exception is better than inspecting up front.

(In general, Python prefers EAFP over LBYL.)

jwt/__init__.py
wbolster: Okay, will change.  I just have a personal distaste for non-module imports. ;)

wbolster:
I've pushed an updated version. The imports are simpler now, and I've updated the README. The exception approach for Python 2.6 compatibility due to missing `timedelta.total_seconds()` is left as-is; see the comments on the previous iteration for details.

jwt/__init__.py
mark-adams: Great point. That works for me. =)

mark-adams:
Looks good to me

jpadilla:
@wbolster thinking about dropping the checks for `total_seconds` and always using:

```python
leeway = leeway.days * 24 * 60 * 60 + leeway.seconds
```

wbolster:
Sure, would be fine with a comment explaining why.  But think about those poor microseconds! :)

PR #64, opened 2015-01-05
Use constant time string comparison routine from hmac module

### LINKED ISSUE #63: Use constant time string comparison routine from the stdlib if available
The summary says it all; PR upcoming.

wbolster:
...and only use the current implementation as a fallback.

Fixes #63.


jpadilla:
@wbolster this is great, thanks!

PR #66, opened 2015-01-06
Verify that decoded header and payload are json objects

### LINKED ISSUE #65: Missing checks that decoded header and payload are actually json objects
Consider this sample evil JWT string: `eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.InN0cmluZyBjb250YWluaW5nIG5iZiBhbmQgZXhwIHRvIGZvb2wgdGhlIFwiaW5cIiBvcGVyYXRvciI.OP64cbMBNsV3_2dAgL_VD96-gAvKksoEMpwIL6BAh3M` using the secret `secret`.

The payload in this JWT is not a JSON encoded object (dict in python), but a string containing the substrings "nbf" and "exp". This fools the validation since the `in` operator works both on strings and dictionaries, resulting in a `TypeError` instead of a `jwt.DecodeError`:

```python
  File "tests/test_jwt.py", line 796, in test_check_issuer
    decoded = jwt.decode(token, 'secret', issuer=issuer)
  File "/Users/wbolster/Projects/pyjwt/jwt/__init__.py", line 324, in decode
    verify_expiration, leeway, **kwargs)
  File "/Users/wbolster/Projects/pyjwt/jwt/__init__.py", line 395, in verify_signature
    if payload['nbf'] > (utc_timestamp + leeway):
TypeError: string indices must be integers
```

The relevant part of the spec (http://self-issued.info/docs/draft-ietf-oauth-json-web-token.html#Validating) mandates that a proper check is done:

> Verify that the resulting octet sequence is a UTF-8 encoded representation of a completely valid JSON object [...]

...but currently pyjwt doesn't do this.

I'm not sure whether this has serious security implications, but at least it allows an attacker (who already knows the secret) to crash the application instead of simply rejecting the invalid JWT.

wbolster:
Fixes #65.

### DIFF
--- a/jwt/__init__.py
+++ b/jwt/__init__.py
@@ -343,6 +343,8 @@ def load(jwt):
         header = json.loads(header_data.decode('utf-8'))
     except ValueError as e:
         raise DecodeError('Invalid header string: %s' % e)
+    if not isinstance(header, Mapping):
+        raise DecodeError('Invalid header string: must be a json object')
 
     try:
         payload_data = base64url_decode(payload_segment)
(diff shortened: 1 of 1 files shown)


jpadilla:
Yeah, this is totally awesome, can't believe it still wasn't in place. Thanks!

Issue #42, opened 2014-10-12
Adding support for Ed25519 signatures


skion:
Would anyone upstream be interested in adding support for NaCl-style [Ed25519](http://ed25519.cr.yp.to/) signatures? I have done some preliminary work [in here](https://github.com/skion/pyjwt/tree/nacl-pure) based on [PyNaCl](https://github.com/pyca/pynacl). At the moment it includes a fallback (and prohibitively slow) pure-Python implementation as well, but I could easily be persuaded to create a PR without that part and/or based on feedback.

wbolster:
@skion see #70 for more discussion re. extensions.

mark-adams:
@jpadilla Since we've decided this is best handled by some form of extensibility (like #70), can we go ahead and close this issue?

skion:
Most definitely; looking forward to trying the registry stuff soon.

stela:
Since January 2017, Ed25519 and Ed448 signatures are now standardized by this RFC: https://tools.ietf.org/html/rfc8037 (using alg=EdDSA)
There's a registry of algorithm definitions to use for JOSE implementations at https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms

PR #71, opened 2015-01-06
Fixes #70. Implemented registry pattern for class-based algorithms

### LINKED ISSUE #70: Move algorithm-specific logic to be class-based to allow for better extensibility
In #42, we discussed changing to more of a registry model for registration of algorithms. This issue is suggesting that we move to that sort of a model.

mark-adams:
Algorithm-specific logic has been moved to the algorithms module and refactored into classes to make it a bit easier for others to add new standard or non-standard algorithms to the library.

Things to note:
* The public APIs (`jwt.encode()` and `jwt.decode()`) are completely unchanged
* `jwt.register_algorithm()` has been added to the public API for extensibility
* Crypto code that formally resided in `signing_methods`, `verify_methods`, and `prepare_key_methods` lists / functions has been rearranged into classes in `algorithms.py`. The sign, verify, and prepare_key methods are all nearly identical to their original versions.

### DIFF
--- a/jwt/__init__.py
+++ b/jwt/__init__.py
@@ -12,6 +12,7 @@ import hmac
 from datetime import datetime, timedelta
 from calendar import timegm
 from collections import Mapping
+from utils import base64url_encode, base64url_decode
 
 from .compat import (json, string_types, text_type, constant_time_compare,
                      timedelta_total_seconds)
@@ -70,24 +71,10 @@ class InvalidAudienceError(InvalidTokenError):
 class InvalidIssuerError(InvalidTokenError):
     pass
 
-
 # Compatibility aliases (deprecated)
 ExpiredSignature = ExpiredSignatureError
 InvalidAudience = InvalidAudienceError
 InvalidIssuer = InvalidIssuerError
-
-def base64url_decode(input):
-    rem = len(input) % 4
-
-    if rem > 0:
-        input += b'=' * (4 - rem)
-
-    return base64.urlsafe_b64decode(input)
-
-
-def base64url_encode(input):
-    return base64.urlsafe_b64encode(input).replace(b'=', b'')
-
 
 def header(jwt):
     if isinstance(jwt, text_type):
--- a/jwt/__init__.py
+++ b/jwt/__init__.py
@@ -6,7 +6,6 @@ http://self-issued.info/docs/draft-jones-json-web-token-01.html
 """
 
 import binascii
-import sys
 
 from calendar import timegm
 from collections import Mapping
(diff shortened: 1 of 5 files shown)


mark-adams:
Agreed. I meant to do that (NotImplementedError) but I was too excited about getting to the rest of the refactoring and then I missed it when I picked this branch back up after the holidays. :-)

I'll push up the commit in just a sec

mark-adams:
There we go. Tests are now green again.

jpadilla:
@progrium I'd like your take on this.

mark-adams:
@jpadilla It's always scary whenever you move around 80% of the code in a library, huh? :-)

wbolster:
This PR needs some rework to apply cleanly on top of current master.

mark-adams:
Thanks for the heads up @wbolster!

@progrium @jpadilla This pull request has been rebased against the master with the PRs from the last 24 hours and is ready for your review.

jpadilla:
@mark-adams @wbolster thanks! Will try to have @progrium check this out to have his feedback since it's a "big" fundamental change before we make any decisions.

mark-adams:
Absolutely!! I totally understand.

 However, I do want to point out two very important things to keep in mind here:
* The public API (`jwt.encode()` and `jwt.decode()`) is completely unchanged with the exception of the fact that register_algorithm has been added as an extensibility point.
* The crypto related code has simply been shifted form using signing_methods, verify_methods, and prepare_key_methods to using an OO approach that is a little more understandable and a little easier to maintain. The contents of the sign, verify, and prepare methods are nearly identical to their original versions in `__init__.py`

A LOT of code looks like it changed here, but a lot of the code actually didn't change here. It's just organized a bit differently. :-)

That being said, I totally understand the need for review. I'll anxiously await @progrium's review. Let me know if y'all have any questions.

mark-adams:
@progrium, This has been rebased against the current master and is ready for your review.

cc: @jpadilla @wbolster

jpadilla:
@mark-adams ready to move this forward. One thing I'd like to do before merging this in would be to write some comment blocks for the algorithms.

mark-adams:
Rebased against the latest master and added some comments for `jwt.algorithms` and `jwt.register_algorithm`

jpadilla:
@mark-adams awesome thanks!

skion:
A bit late to the party, but I'm trying this in view of #42 at the moment...

The Algorithm abstraction doesn't provide access to the unprotected header, just the payload. This can be inconvenient if you want the algorithm to manage e.g. the `kid` header automatically.

Problem is that since the `jwt.header()` method has been removed, you also [can't work around it on the application level](https://github.com/jpadilla/pyjwt/pull/85#issuecomment-100196275).

Any tips?

mark-adams:
In the 2.0 release (#144 to be specific), we're planning on adding additional support for JWK that should make key management a bit easier without having to resort to a lot of trickery. I think this will help resolve a lot of what you're dealing with.

I wouldn't be opposed to adding a function to retrieve the unverified header values. Created #155

PR #72, opened 2015-01-06
Invalid validation of audience claim

wbolster:
The spec at http://self-issued.info/docs/draft-ietf-oauth-json-web-token.html#audDef states:

> The aud (audience) claim identifies the recipients that the JWT is intended for. Each principal intended to process the JWT MUST identify itself with a value in the audience claim. If the principal processing the claim does not identify itself with a value in the aud claim when this claim is present, then the JWT MUST be rejected.

As I understand it, Pyjwt doesn't seem to implement this requirement correctly. Basically it reverses the checks. Let me explain.

Pyjwt only checks the audience claim when an `audience=...` arg was specified to `decode()`, meaning that any `audience` claim in the JWT is blindly ignored if no such arg was passed. This results in the JWT bein  accepted, instead of being rejected. (Security problem!)

Furthermore, the `decode(audience=...`) arg allows both a string and a list of strings, but the audience claim in the JWT payload is always assumed to be a single string, which is against the spec:

> In the general case, the aud value is an array of case-sensitive strings, each containing a StringOrURI value. In the special case when the JWT has one audience, the aud value MAY be a single case-sensitive string containing a StringOrURI value. The interpretation of audience values is generally application specific. Use of this claim is OPTIONAL.

Offending code:

```python
    if audience is not None:
        if isinstance(audience, list):
            audiences = audience
        else:
            audiences = [audience]

        if payload.get('aud') not in audiences:
            raise InvalidAudience('Invalid audience')
```

Hopefully you understand what I meant when I said pyjwt "reverses" the checks.

Let me know your thoughts; I can come up with a PR when we agree on the correct approach and understanding of the spec.

### DIFF
--- a/jwt/__init__.py
+++ b/jwt/__init__.py
@@ -398,6 +398,9 @@ def verify_signature(payload, signing_input, header, signature, key='',
         else:
             leeway = leeway.total_seconds()
 
+    if not isinstance(audience, (basestring, type(None))):
+        raise TypeError('audience must be a string or None')
+
     try:
         algorithm = header['alg'].upper()
         key = prepare_key_methods[algorithm](key)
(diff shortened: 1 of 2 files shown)


wbolster:
@jpadilla @mark-adams I've changed the verification logic. Please review.

The intention of my code is almost what @mark-adams wrote, but not quite. In order for validation to pass:

* 'aud' must not be present in the claimset _and the application must not specify `audience=` either_
* if 'aud' is a list, it must contain the application's value
* if 'aud' is a string, it must match the application's value

Any other formats (e.g. a number or a nested object) for the 'aud' field are also considered invalid.

The current implementation _does not_ support multiple audience values specified by the application. If this is desired, I can add it as well. The only use case I can come up with for this though is a service that handles authorization for multiple audiences, which is a bit far-fetched perhaps. Thoughts?

wbolster:
Btw, this PR is on top of current master, so it uses the "old" exception names. I will rebase (and update) once you have merged #60.

wbolster:
Okay, rebased onto current master, and added an additional string check.

jpadilla:
@wbolster awesome, thanks!

Issue #67, opened 2015-01-06
Handle all compatibility branching in a compat.py


jpadilla:
I'd like to add a `compat.py` to handle all compatibility branching in one place, removing any of that clutter from the actual code.

jpadilla:
- prepare_key_methods
- signing_methods
- verify_methods
- constant_time_compare

Perhaps also handling of `.total_seconds()` should also go here.

wbolster:
While you're at it, drop usage of `unicode_literals` as well. :) https://github.com/PythonCharmers/python-future/issues/22

PR #74, opened 2015-01-07
Handle compatibility branching in a compat module

jpadilla:
Ref #67

### DIFF
--- a/jwt/compat.py
+++ b/jwt/compat.py
@@ -1,6 +1,6 @@
 """
 The `compat` module provides support for backwards compatibility with older
-versions of django/python, and compatibility wrappers around optional packages.
+versions of python, and compatibility wrappers around optional packages.
 """
 # flake8: noqa
 import sys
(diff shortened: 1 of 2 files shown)


jpadilla:
@wbolster @mark-adams would like your thoughts on this.

jpadilla:
@wbolster @mark-adams will probably merge this if there aren't no objections and release a new version with all your awesome work.

wbolster:
Sure, next step is to move all actual code out of `__init.py__`, and only import public API there. :-)

jwt/compat.py
```diff
@@ -0,0 +1,59 @@
...
+try:
+    import json
+except ImportError:
+    import simplejson as json
+
+
+if sys.version_info >= (3, 0, 0):
+    unicode = str
+    basestring = str
+else:
```
wbolster: drop this branch, it doesn't do anything

jwt/compat.py
```diff
@@ -0,0 +1,59 @@
...
+import sys
+import hmac
+
+try:
+    import json
+except ImportError:
+    import simplejson as json
+
+
+if sys.version_info >= (3, 0, 0):
```
wbolster: make this a PY3 constant (like six), and use it elsewhere as well

mark-adams:
I know this is already merged, but I just wanted to say "Looks great!" :-)

jwt/compat.py
jpadilla: @wbolster thanks, borrowed what six was doing for this.

jpadilla:
@wbolster @mark-adams thanks again.

@wbolster RE moving the actual code out of __init__, on that right now. Will open a PR in a few minutes.

mark-adams:
Interesting note regaridng moving code out of init: A lot of this has been done in #71 :-)

jpadilla:
@mark-adams true that, will hold off on this until then.

jpadilla:
Just released [v0.4.1](https://github.com/progrium/pyjwt/releases/tag/0.4.1).

PR #75, opened 2015-01-11
Remove unicode_literals

jpadilla:


wbolster:
This is not complete; #76 is.  This PR can be closed since #76 contains the same change.

PR #77, opened 2015-01-12
Be politically correct

wbolster:


jpadilla:
Yeah, good catch.

wbolster:
i prefer to use the word 'polizei' in cases like this ;)

PR #78, opened 2015-01-18
Cleanup __init__ module

jpadilla:
- Exposes public API
- Moves exceptions to separate module
- Uses relative imports
- Fixes tests that used private API

@mark-adams @wbolster Remembered we talked at some point of doing this. Thoughts?


jwt/api.py
```diff
@@ -0,0 +1,196 @@
+import binascii
+
+from calendar import timegm
+from collections import Mapping
+from datetime import datetime, timedelta
+
+from .utils import base64url_decode, base64url_encode
```
mark-adams: Super minor, but I'd flip around the .utils and .compat imports so they're alphabetical.

mark-adams:
Looks good to me

jpadilla:
@mark-adams aiming to release and tag a v1 stable version soon.

PR #79, opened 2015-01-18
Refactored package information so it only has to be in one place

mark-adams:
When I saw that the package info was added to __init__.py, I wanted to try and find a way to make it where __init__.py and setup.py used the same source for their information so we only have to change it once anytime that we update anything.

With that in mind, I created __about__.py which contains the package, version, and author information and changed __init__.py and setup.py to pull from there. As a bonus, this let me remove the `get_version()` regex-based function that was previously used in setup.py

cc: @jpadilla

### DIFF
--- a/jwt/__about__.py
+++ b/jwt/__about__.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+
+__all__ = [
+    "__title__", "__description__", "__url__", "__version__", "__author__",
+    "__email__", "__license__", "__copyright__", "__description__"
+]
+
+__title__ = 'pyjwt'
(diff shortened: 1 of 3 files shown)


jpadilla:
@mark-adams hey thanks for this. For now I'd prefer to keep it as is which I think is what most people are used to and other well maintained packages do. What I will do is just use a few things like `__version__` from the package which I saw [worked](https://github.com/kennethreitz/requests/blob/master/setup.py).

mark-adams:
No problem. I just saw [another popular project](https://github.com/pyca/cryptography/blob/master/setup.py) and thought it was a good idea.

PR #85, opened 2015-01-25
Removed `api.header`

mark-adams:
The header function isn't a public part of the API, is not covered by any tests, and really doesn't seem to serve a practical purpose so I propose removing it.


jpadilla:
@mark-adams yeah let's get rid of it. Thanks!

skion:
Hmm, what is now the suggested way to access the unprotected header before calling `decode()`?

My use case is that I need access to a `kid` header which is in turn used as a parameter to decode.

mark-adams:
@skion See #155

Issue #83, opened 2015-01-22
Can't decode tokens encoded with the "none" algorithm


seedifferently:
There seems to be some problems with decoding tokens using the "none" algorithm:

    >>> jwt.encode({'some': 'payload'}, 'secret', 'none')
    b'eyJ0eXAiOiJKV1QiLCJhbGciOiJub25lIn0.eyJzb21lIjoicGF5bG9hZCJ9.'
    >>> jwt.decode('eyJ0eXAiOiJKV1QiLCJhbGciOiJub25lIn0.eyJzb21lIjoicGF5bG9hZCJ9.', 'secret')
    [...]
    KeyError: 'NONE'
    [...]
    DecodeError: Algorithm not supported

jpadilla:
I've never understood why we even have that option in the first place. But it's meant to be used like this:

```python
jwt_message = jwt.encode({'some': 'payload'}, key=None, algorithm=None)
jwt.decode(jwt_message, verify=False)
```

seedifferently:
@jpadilla ah, thanks. Sorry about the noise.

PR #86, opened 2015-01-25
Fix alg header parameter case sensitivity

mark-adams:
The [JWS spec](https://tools.ietf.org/html/draft-ietf-jose-json-web-signature-41#section-4.1.1) says: "The "alg" value is a case-sensitive ASCII string containing a StringOrURI value."

Our implementation currently normalizes case of the alg by uppercasing on decode. This PR resolves that behavior so that, by default, "none" is valid and "NONE" is not. Tests are added as well.

This PR also fixes a small bug with signature verification on the NoneAlgorithm related to the fact the `verify()` method is missing a parameter (whoops!)

### DIFF
--- a/jwt/algorithms.py
+++ b/jwt/algorithms.py
@@ -74,7 +74,7 @@ class NoneAlgorithm(Algorithm):
     def sign(self, msg, key):
         return b''
 
-    def verify(self, msg, key):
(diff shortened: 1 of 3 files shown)


jpadilla:
Thanks! This explains the odd traceback on #83.

PR #87, opened 2015-01-25
Improve code coverage by adding a ton of tests

mark-adams:
Note: Please review PR #86 prior to this one. (It makes the assumption that #86 is merged and if it is not; it will need to be updated.

This PR makes several changes primarily focused once again around improving testing:
 * Added a ton of tests to cover areas that weren't covered before
 * Added a pep8 tox env to run pep8 tests as part of the build: `tox -e pep8`
 * Moved keys into tests/keys to make the directory a bit cleaner
 * Added test_algorithms.py for algorithms module tests
 * Changed setup.py to run all tests in tests/ (instead of just test_jwt.py)
 * Added new environments for py34-nocrypto and py27-nocrypto to help cover code where we don't have crypto libraries installed.
 * Removed `json` from `compat.py` since [anything >= 2.6 supports json](https://docs.python.org/2/library/json.html)


mark-adams:
Boom! 100% code coverage!

jpadilla:
You sir, have outdone yourself once again! I'm confident we're almost ready for that v1.0.0 release thanks to all this awsome work.

Issue #81, opened 2015-01-21
Encoding with RSA public key and decoding with private key


svnv:
In the documentation it says:
> When using the RSASSA-PKCS1-v1_5 algorithms, the key argument in both jwt.encode() and jwt.decode() ("secret" in the examples) is expected to be either an RSA public or private key in PEM format.

When i try to encode the payload with the public key i get an error:

    >>> jwt_message = jwt.encode(payload, pub_rsakey, 'RS384')
    Traceback (most recent call last):
      File "<input>", line 1, in <module>
      File "/Users/sven/code/sso/env/lib/python2.7/site-packages/jwt/__init__.py",
     line 295, in encode
        signature = signing_methods[algorithm](signing_input, key)
      File "/Users/sven/code/sso/env/lib/python2.7/site-packages/jwt/__init__.py",
     line 131, in <lambda>
        'RS384': lambda msg, key: sign_rsa(msg, key, hashes.SHA384()),
      File "/Users/sven/code/sso/env/lib/python2.7/site-packages/jwt/__init__.py",
     line 106, in sign_rsa
        signer = key.signer(
    AttributeError: '_RSAPublicKey' object has no attribute 'signer'

What is happening is that the implementation in [algorithms.py, line 137](https://github.com/jpadilla/pyjwt/blob/master/jwt/algorithms.py#L137) tries to sign with the public key, but the `cryptography.hazmat.backends.openssl.rsa._RSAPublicKey` has no `signer` method: https://cryptography.io/en/latest/hazmat/primitives/interfaces/#cryptography.hazmat.primitives.interfaces.RSAPublicKey

svnv:
So, it look like I misunderstood how jwt works. I somehow thought that the payload was also encrypted with the RSA key. (I edited my previous post in light of my new knowledge.)

The documentation should at least be updated to reflect that encoding (signing) is only possible with a private RSA key and decoding (verification) only works with a private key.

jpadilla:
@svnv a pull request is welcome with your proposal to what the documentation should say about this.

mark-adams:
I think there is a general assumption that users of JWT understand the basics of asymmetric crypto and how a digital signature works. However, it might not hurt to provide a link to [Wikipedia's article on Public-key crypto](http://en.wikipedia.org/wiki/Public-key_cryptography)

jpadilla:
:+1: sounds good to me

jpadilla:
Closed via #88

PR #88, opened 2015-01-25
Updated documentation to match the latest version of the JWT specification (draft 32)

### LINKED ISSUE #80: Update documentation to reference the latest spec
Apparently there have been 30+ iterations to the JWT spec since the pyjwt documentation was created. Do we need to update the docs to reflect that we're up to speed with the latest versions of the spec?

https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32

@jpadilla Thoughts?

mark-adams:
The title says it all. Fixes #80 and addresses the concerns from #81.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,13 +79,18 @@ Change the algorithm with by setting it in encode:
 jwt.encode({'some': 'payload'}, 'secret', 'HS512')
 ```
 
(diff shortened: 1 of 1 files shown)


jpadilla:
Boom! Thanks!

PR #100, opened 2015-03-15
Resolved #93 by adding a section the README detailing how to extract public / private keys from an x509 certificate.

### LINKED ISSUE #93: Decoding RS256 JWT
I am trying to decode a JWT encoded with `RS256`.

When I run the following:
```
secret = Crypto.PublicKey.RSA.importKey(PUBLIC_KEY)
```
I get:
```
ValueError: RSA key format is not supported
```

I suspect something might be wrong with the public key format:
```
PUBLIC_KEY= """-----BEGIN CERTIFICATE-----
MIIDETCCAfmgAwIBAgI....sRPdPR8
-----END CERTIFICATE-----"""
```

mark-adams:
* Added an FAQ section to the README with some text explaining how to extract public / private keys from a x509 certificate since we seem to get this question a lot.
* Moved "Tests" higher up so that Installation, Usage, and Tests are all grouped together.
* Wrapped a couple long lines in the existing README

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,6 +247,22 @@ decoded = jwt.decode(token, 'secret', audience='urn:foo')
 
 If the audience claim is incorrect, `jwt.InvalidAudienceError` will be raised.
 
+## Frequently Asked Questions
+
+**How can I extract a public / private key from a x509 certificate?**
+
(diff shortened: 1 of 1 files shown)


jpadilla:
Great work!

Issue #99, opened 2015-03-14
Support for App Engine


garyp:
Google App Engine's python runtime environment unfortunately only supports the "pycrypto" library and not the newer "cryptography" library (they have a pre-vetted list of native libraries that are supported): https://cloud.google.com/appengine/docs/python/tools/libraries27. I'm using the 0.3.x branch of pyjwt on App Engine right now and it'd be nice to be able to ugprade to 0.4.x.

The only solution I can think of is to bring back the old pycrypto code and use it as a fallback when the "cryptography" library isn't available. If you're ok with continuing to support pycrypto within pyjwt, I don't mind putting together a pull request to implement the fallback logic.

jpadilla:
GAE strikes again! Are you using RSA or ECDSA signatures? If not you wouldn't need `pycypto` or `cryptography` at all. I think we could definitely evaluate the possibilities of abstracting both libraries, and support both of them without including too much about each in our implementation. Does that make sense?

garyp:
I'm using RS256 signatures. Yeah, abstracting both libraries is what I was thinking. The code already seems pretty well abstracted, so it doesn't seem like it'd be a lot of work to define a common interface.

jpadilla:
@garyp I'm down to seeing a proposal for this.

garyp:
@jpadilla Great. I'll probably work on this in a week or two.

mark-adams:
This can now be accomplished by using `register_algorithm` and the and the jwt.contrib.algorithms.pycrypto classes.

johnlockwood-wf:
@mark-adams, how do you use `register_algorithm` with the `jwt.contrib` algorithms? When I `python setup.py install`, there is no contrib package in jwt.

johnlockwood-wf:
this seems to be an issue with the setup.py. It needs to list all the packages like:

    setup(
        name='PyJWT',
        ...
        packages=['jwt', 'jwt.contrib', 'jwt.contrib.algorithms'],

PR #103, opened 2015-03-16
Back from the dead: PyCrypto and ECDSA

mark-adams:
This PR takes care of a chunk of the work towards resolving #99. Specifically, it accomplishes two things:
- Support for PyCrypto-based RSA and ecdsa-based EC signatures is restored in the `jwt.contrib.algorithms` module
- Each of the preferred `jwt.algorithms` as well as the `jwt.contrib.algorithms` now have static class variables for SHA256, SHA384, and SHA512 to make our default registration code a little simpler (it no longer has to know the specifics about the hash objects). This also could eventually allow us to write some shared test code between the preferred and `jwt.contrib.algorithms` algorithms.

This PR would allow a consumer to call register_algorithm and pass in the `RSAAlgorithm` or the `ECAlgorithm` from `jwt.contrib.algorithms` if they would like to use the legacy libraries. I think we could make our default algorithms code much more complex by trying to autodetect all three libraries (cryptography, pycrypto, and ecdsa) but I think it is much simpler for us to only detect for cryptography and simply make the other algorithms available for those on AppEngine or other environments where they need specific support.

### DIFF
--- a/jwt/algorithms.py
+++ b/jwt/algorithms.py
@@ -23,18 +23,18 @@ def _register_default_algorithms():
     Registers the algorithms that are implemented by the library.
     """
     register_algorithm('none', NoneAlgorithm())
-    register_algorithm('HS256', HMACAlgorithm(hashlib.sha256))
-    register_algorithm('HS384', HMACAlgorithm(hashlib.sha384))
-    register_algorithm('HS512', HMACAlgorithm(hashlib.sha512))
+    register_algorithm('HS256', HMACAlgorithm(HMACAlgorithm.SHA256))
+    register_algorithm('HS384', HMACAlgorithm(HMACAlgorithm.SHA384))
+    register_algorithm('HS512', HMACAlgorithm(HMACAlgorithm.SHA512))
 
     if has_crypto:
-        register_algorithm('RS256', RSAAlgorithm(hashes.SHA256()))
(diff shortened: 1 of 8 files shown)


jpadilla:
@mark-adams dude this is pure gold right here!

jwt/algorithms.py
```diff
@@ -86,6 +86,8 @@ class HMACAlgorithm(Algorithm):
     def __init__(self, hash_alg):
         self.hash_alg = hash_alg
 
+    SHA256, SHA384, SHA512 = hashlib.sha256, hashlib.sha384, hashlib.sha512
```
jpadilla: I think I'd prefer to change the style of these everywhere to separate lines and before the `__init__` method. I think it's easier to parse.

```python
class HMACAlgorithm(Algorithm):
    """
    Performs signing and verification operations using HMAC
    and the specified hash function.
    """
    SHA256 = hashlib.sha256
    SHA384 = hashlib.sha384
    SHA512 = hashlib.sha512

    def __init__(self, hash_alg):
        self.hash_alg = hash_alg
```

mark-adams:
@jpadilla Thanks!! I think you're right about the hash function variables. I debated which way to lay them out for a while and was torn between my way and the way you suggested. You're probably right, it is more readable your way. I went ahead and made the change.

PR #109, opened 2015-03-17
Added some fixes related to algorithm and key choice

mark-adams:
Added validations for the following:
* Asymmetric keys and x509 certificates cannot be used as HMAC keys
* `encode()` and `decode()` should not accept a value for key if the algorithm is "none"


jwt/algorithms.py
```diff
@@ -96,6 +103,11 @@ def prepare_key(self, key):
         if isinstance(key, text_type):
             key = key.encode('utf-8')
 
+        if (b'-----BEGIN PUBLIC KEY-----' in key or b'-----BEGIN CERTIFICATE-----' in key):
+            raise InvalidAlgorithmError(
```
jpadilla: Should this be `InvalidKeyError` or `InvalidAlgorithmError`?

mark-adams: It should be InvalidKeyError... changing now

mark-adams:
FYI, I did some performance tests (500,000 `encode()`s) on HMAC before and after the change (invalid_strings check for public key and certs).

Without new checks: 32.63310008600092
With new checks: 34.56768506300068

Looks like this change slows down HMAC encode by about 6%.

jwt/algorithms.py
```diff
@@ -96,6 +103,17 @@ def prepare_key(self, key):
         if isinstance(key, text_type):
             key = key.encode('utf-8')
 
+        invalid_strings = [
+            b'-----BEGIN PUBLIC KEY-----',
+            b'-----BEGIN CERTIFICATE-----',
+            b'ssh-rsa'
```
sullivanmatt: I'm aware this PR has already been merged, but I wanted to open a brief discussion.

Another possible header is `-----BEGIN RSA PUBLIC KEY-----`, and there may be others.  Is there a more sound way of dealing with this that's somewhat more future-proof?  If the check looked like this, would it be too broad?
```
invalid_strings = [
            b'-----',
            b'ssh-rsa'
]
```

PR #119, opened 2015-03-27
Fixed some references to avoid DeprecationWarnings from cryptography.

mark-adams:
This PR changes around some imports to avoid DeprecationWarnings from cryptography.


mark-adams:
Basically I was dumb and tried to fix something that I already fixed.

PR #124, opened 2015-03-29
Refactor tests to remove references to private API methods

mark-adams:
One of the biggest advantages to tests is that they allow us to test the public APIs of the library independent of the underlying implementation (private APIs). This allows us to completely change the underlying implementation while having confidence that our library still works.

This PR changes all of our tests that use private APIs and methods and refactors them to focus on public APIs. All of the code is still being tested; its just being tests through only public APIs now.

**NOTE: Please review #122 before this one. (This PR is based off of #122 and will be much less confusing once it is merged)


jpadilla:
Oh yea! Way easier to understand tests now.

mark-adams:
Thanks!! I was about to try and make a proposal on reworking some of the internal APIs and it would be WAAAAY easier if the tests only focused on the public ones.

PR #125, opened 2015-03-29
Moved claims validation into a separate private method to make it simpler to understand the code.

mark-adams:
Moved claims validation into a separate private method to make it simpler to understand the code.


tests/test_api.py
```diff
@@ -479,9 +479,6 @@ def test_decode_with_expiration_with_leeway(self):
         for leeway in (3, timedelta(seconds=3)):
             self.jwt.decode(jwt_message, secret, leeway=leeway)
 
-            self.jwt._verify_signature(decoded_payload, signing, header,
```
mark-adams: I missed this test when I removed private API tests earlier.

jpadilla:
LGTM

PR #126, opened 2015-03-29
Add check for iat and new exception for nbf claims

mark-adams:
* Added checks on iat to make sure that a token can't be issued for the future
* Changed nbf exception to ImmatureSignatureError (matches ruby-jwt)

### DIFF
--- a/jwt/__init__.py
+++ b/jwt/__init__.py
@@ -20,7 +20,7 @@ from .api import (
     encode, decode, register_algorithm, unregister_algorithm, PyJWT
 )
 from .exceptions import (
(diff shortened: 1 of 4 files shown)


jpadilla:
LGTM. By the way, should we go ahead and add support for those missing "checks" on http://jwt.io/?

mark-adams:
I was just thinking the exact same thing, but I don't want to have a million arguments that have to be passed into `decode()`. I was thinking about doing something like what the Ruby library does and passing in a dict of options. I'll create an issue to discuss further.

PR #128, opened 2015-04-02
Fixed command line encoding

nielsdraaisma:
I tried running the example invocation 
jwt --key=secret iss=me exp=1302049071

This failed and returned an error "Not enough segments" followed by a sys.exit(1).
It looks like it tried to parse the encoding parameters and JWT token data, which obviously fails.

I've worked around it and the example scenario's work,but may be a --encode/--decode parameter would be more explicit.

### DIFF
--- a/bin/jwt
+++ b/bin/jwt
@@ -71,21 +71,22 @@ The exp key is special and can take an offset to current Unix time.
     options, arguments = p.parse_args()
 
     if len(arguments) > 0 or not sys.stdin.isatty():
-        # Try to decode
-        try:
-            if not sys.stdin.isatty():
-                token = sys.stdin.read()
-            else:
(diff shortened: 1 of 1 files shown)


jpadilla:
Ugh this thing. I think it either needs to go or get rebuilt. Plus it needs some kind of tests.

@nielsdraaisma do you use it for anything significant/important?

nielsdraaisma:
Don't use it at all, came across it looking for something that could generate a token, and that was broken.

mark-adams:
@jpadilla Are we wanting to keep this? Isit useful

jpadilla:
@mark-adams blah I'd be happier without it really. What would the best "semver" way of getting rid of it soon?

Issue #127, opened 2015-03-29
Add more flexible and complete verification options


mark-adams:
I was thinking that it might be useful for us to implement more flexible verification options.

I propose something like this:

```
def decode(token, secret, options=None)
```

where options is a dict that looks something like this:
```
options = {
   'verify_signature': True,
   'verify_exp': True,
   'verify_nbf': True,
   'verify_iat': True,
   'verify_aud`: True
}
```

This is similar to what [ruby-jwt does](https://github.com/progrium/ruby-jwt/blob/master/lib/jwt.rb#L110)
We could make it where options could be specified globally (at the PyJWT object level) or as an override argument to `decode()`

jpadilla:
@mark-adams yeah I can see how this can be useful. For example that #129 use case could just set `verify_aud` to False.

mark-adams:
I'll take a look

michaeldavis-wf:
@mark-adams I know timelines are to be taken with a grain of salt, but do you have a guesstimate on how long 1.1.0 might be?  I would love to be able to pull in the changes from this PR.

If it is going to be too long, would it be possible to cut a 1.0.2 release with the current master?

mark-adams:
Moved this convo to #137

PR #129, opened 2015-04-03
Not specifying an audience to decode() causes an InvalidAudienceError

michaeldavis-wf:
When I attempt to decode an access token that contains an `'aud'` claim without passing an audience into the decode method, the `_verify_claims()` method raises an `InvalidAudienceError`.

There wasn't a test that covered this case, so I'm not sure if it is intended behavior or not.  If it is, feel free to ignore me.  I can work around it easily enough.  If it isn't, I made the change and added a test to cover it.

#### Example:
```
>>> j = jwt.encode({'test':'something','aud':'test'}, 'secret')
>>> jwt.decode(j, 'secret')
  File "/Users/michael/.virtualenvs/webfilings/wf-oauthtwo-pyserver/lib/python2.7/site-packages/jwt/api.py", line 117, in decode
    key, algorithms, **kwargs)
  File "/Users/michael/.virtualenvs/webfilings/wf-oauthtwo-pyserver/lib/python2.7/site-packages/jwt/api.py", line 208, in _verify_signature
    # verified since the token does not contain a claim.
jwt.exceptions.InvalidAudienceError: Invalid audience
```

### DIFF
--- a/jwt/api.py
+++ b/jwt/api.py
@@ -214,7 +214,12 @@ class PyJWT(object):
             if exp < (now - leeway):
                 raise ExpiredSignatureError('Signature has expired')
 
-        if 'aud' in payload:
+        if audience is not None:
+            # Application specified an audience, but it could not be
+            # verified since the token does not contain a claim.
+            if 'aud' not in payload:
+                raise InvalidAudienceError('No audience claim in token')
+
             audience_claims = payload['aud']
             if isinstance(audience_claims, string_types):
                 audience_claims = [audience_claims]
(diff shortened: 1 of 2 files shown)


mark-adams:
This is intended behavior. If `aud` claim is present, it indicates that the creator of the JWT intended it to be for a specific audience / purpose. If the application consuming the JWT isn't that audience, than the JWT should be considered invalid.

michaeldavis-wf:
I would agree that the application consuming the JWT should be the intended audience, but I believe there are use cases where the verifying application and the consuming application are separate applications.

We are running into this issue providing an endpoint for token validation.  It is based off of Google's OAuth 2.0 [tokeninfo endpoint.](https://developers.google.com/accounts/docs/OAuth2UserAgent#validatetoken).  

We take in the token, validate the signature and expiration, and return the decoded token.  In the future, this will also support a token revocation list.  The audience is still being checked by the appropriate actor, but the actual decoding of the token is done by the service.  Google even includes a note that the audience still needs to be checked by the client.

We also intend to run an edge service that will validate every token into a set of applications.  The edge service would perform the same sort of signature and expiration validation that the current tokeninfo endpoint provides, but the audience validation would be left to the consuming application.

Let me know what you think.

mark-adams:
I'm going to reject this in favor of us working towards resolving #127

PR #131, opened 2015-04-07
Add flexible and complete verification options

### LINKED ISSUE #127: Add more flexible and complete verification options

michaeldavis-wf:
This adds the ability to use a more flexible option dictionary.  The options dictionary can be specified at the PyJWT object level or as an override passed directly into `decode()`

Attempts to fix #127

### DIFF
--- a/jwt/api.py
+++ b/jwt/api.py
@@ -16,7 +16,7 @@ from .utils import base64url_decode, base64url_encode
 
 
 class PyJWT(object):
-    def __init__(self, algorithms=None):
+    def __init__(self, algorithms=None, options=None):
         self._algorithms = get_default_algorithms()
         self._valid_algs = set(algorithms) if algorithms is not None else set(self._algorithms)
 
@@ -24,6 +24,22 @@ class PyJWT(object):
         for key in list(self._algorithms.keys()):
             if key not in self._valid_algs:
                 del self._algorithms[key]
+
+        if not options:
+            options = {}
+
+        self.default_options = {
+            'verify_signature': True,
+            'verify_exp': True,
+            'verify_nbf': True,
+            'verify_iat': True,
+            'verify_aud': True,
+        }
+
+        try:
+            self.options = {k: options[k] if k in options else v for k, v in self.default_options.items()}
+        except (ValueError, TypeError) as e:
+            raise TypeError('options must be a dictionary: %s' % e)
 
     def register_algorithm(self, alg_id, alg_obj):
         """
@@ -110,14 +126,16 @@ class PyJWT(object):
 
         return b'.'.join(segments)
 
-    def decode(self, jwt, key='', verify=True, algorithms=None, **kwargs):
+    def decode(self, jwt, key='', verify=True, algorithms=None, options=None, **kwargs):
         payload, signing_input, header, signature = self._load(jwt)
 
         if verify:
-            self._verify_signature(payload, signing_input, header, signature,
-                                   key, algorithms)
+            merged_options = self._merge_options(override_options=options)
+            if merged_options.get('verify_signature'):
+                self._verify_signature(payload, signing_input, header, signature,
+                                       key, algorithms)
 
-            self._validate_claims(payload, **kwargs)
+            self._validate_claims(payload, options=merged_options, **kwargs)
 
         return payload
 
@@ -177,8 +195,8 @@ class PyJWT(object):
         except KeyError:
             raise InvalidAlgorithmError('Algorithm not supported')
 
-    def _validate_claims(self, payload, verify_expiration=True, leeway=0,
-                         audience=None, issuer=None):
+    def _validate_claims(self, payload, audience=None, issuer=None, leeway=0,
+                         options=None, **kwargs):
         if isinstance(leeway, timedelta):
             leeway = timedelta_total_seconds(leeway)
 
@@ -187,7 +205,7 @@ class PyJWT(object):
 
         now = timegm(datetime.utcnow().utctimetuple())
 
-        if 'iat' in payload:
+        if 'iat' in payload and options.get('verify_iat'):
             try:
                 iat = int(payload['iat'])
             except ValueError:
@@ -196,7 +214,7 @@ class PyJWT(object):
             if iat > (now + leeway):
                 raise InvalidIssuedAtError('Issued At claim (iat) cannot be in the future.')
 
-        if 'nbf' in payload and verify_expiration:
+        if 'nbf' in payload and options.get('verify_nbf'):
             try:
                 nbf = int(payload['nbf'])
             except ValueError:
@@ -205,7 +223,7 @@ class PyJWT(object):
             if nbf > (now + leeway):
                 raise ImmatureSignatureError('The token is not yet valid (nbf)')
 
-        if 'exp' in payload and verify_expiration:
+        if 'exp' in payload and options.get('verify_exp'):
             try:
                 exp = int(payload['exp'])
             except ValueError:
@@ -214,7 +232,7 @@ class PyJWT(object):
             if exp < (now - leeway):
                 raise ExpiredSignatureError('Signature has expired')
 
-        if 'aud' in payload:
+        if 'aud' in payload and options.get('verify_aud'):
             audience_claims = payload['aud']
             if isinstance(audience_claims, string_types):
                 audience_claims = [audience_claims]
@@ -232,6 +250,15 @@ class PyJWT(object):
         if issuer is not None:
             if payload.get('iss') != issuer:
                 raise InvalidIssuerError('Invalid issuer')
+
+    def _merge_options(self, override_options=None):
+        if not override_options:
+            override_options = {}
+        try:
+            options = {k: override_options[k] if k in override_options else v for k, v in self.options.items()}
+        except (ValueError, TypeError) as e:
+            raise TypeError('options must be a dictionary: %s' % e)
+        return options
 
 
 _jwt_global_obj = PyJWT()
--- a/jwt/api.py
+++ b/jwt/api.py
@@ -37,8 +37,11 @@ class PyJWT(object):
         }
 
         try:
-            self.options = {k: options[k] if k in options else v for k, v in self.default_options.items()}
-        except (ValueError, TypeError) as e:
+            merged_options = dict()
+            for k, v in self.default_options.items():
+                merged_options[k] = options.get(k, v)
+            self.options = merged_options
+        except AttributeError as e:
             raise TypeError('options must be a dictionary: %s' % e)
 
     def register_algorithm(self, alg_id, alg_obj):
@@ -254,11 +257,15 @@ class PyJWT(object):
     def _merge_options(self, override_options=None):
         if not override_options:
             override_options = {}
+
         try:
-            options = {k: override_options[k] if k in override_options else v for k, v in self.options.items()}
-        except (ValueError, TypeError) as e:
+            merged_options = dict()
+            for k, v in self.options.items():
+                merged_options[k] = override_options.get(k, v)
+        except AttributeError as e:
             raise TypeError('options must be a dictionary: %s' % e)
-        return options
+
+        return merged_options
 
 
 _jwt_global_obj = PyJWT()
--- a/jwt/api.py
+++ b/jwt/api.py
@@ -36,13 +36,7 @@ class PyJWT(object):
             'verify_aud': True,
         }
 
-        try:
-            merged_options = dict()
-            for k, v in self.default_options.items():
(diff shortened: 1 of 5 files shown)


jpadilla:
Oh yeah, this looks good to me.

@mark-adams would like your feedback, since you created #127.

jwt/api.py
```diff
@@ -25,6 +25,25 @@ def __init__(self, algorithms=None):
...
 
+        if not options:
+            options = {}
+
+        self.default_options = {
+            'verify_signature': True,
+            'verify_exp': True,
+            'verify_nbf': True,
+            'verify_iat': True,
+            'verify_aud': True,
```
johnlockwood-wf: I would like to see these abbreviations expanded. The only reason they are abbreviated in the standard is to reduce the amount of data transmitted. Here it just serves to obfuscate.

johnlockwood-wf: How many people will choose to solve, non-verification messages by figuring out they can set all of these to False rather than taking the time to understand what it all means?

jwt/api.py
```diff
@@ -110,14 +129,16 @@ def encode(self, payload, key, algorithm='HS256', headers=None, json_encoder=Non
...
+    def decode(self, jwt, key='', verify=True, algorithms=None, options=None, **kwargs):
         payload, signing_input, header, signature = self._load(jwt)
 
         if verify:
-            self._verify_signature(payload, signing_input, header, signature,
-                                   key, algorithms)
+            merged_options = self._merge_options(override_options=options)
+            if merged_options.get('verify_signature'):
+                self._verify_signature(payload, signing_input, header, signature,
+                                       key, algorithms)
```
johnlockwood-wf: I think one argument that cancel's out the other is dangerous and adds a lot of complexity. It gives attackers an attack vector they could drive a truck through.

johnlockwood-wf: let's make it complicated to setup a non-standard set of claims verifications.

jwt/api.py
```diff
@@ -110,14 +129,16 @@ def encode(self, payload, key, algorithm='HS256', headers=None, json_encoder=Non
...
 
-            self._validate_claims(payload, **kwargs)
+            self._validate_claims(payload, options=merged_options, **kwargs)
```
johnlockwood-wf: What if the call that happens here is configurable? If you build your own claim verifier, you can configure jwt with it.

jwt/api.py
michaeldavis-wf: The only way to avoid consumers skipping verification is to completely disallow it, which ignores a huge number of use cases that need to turn off various checks.

jwt/api.py
michaeldavis-wf: I see no need to over complicate non-standard claims verifications.  I believe this would lead more people not understanding what they are doing, not fewer.

michaeldavis-wf: I can see the argument about verify vs verify_options in the options dict though.

jwt/api.py
michaeldavis-wf: The number one way people mess up security is by trying to write it themselves.  Forcing developers to write their own entire claim verifier, instead of allowing them to turn off the specific functionality that they do not need is going to cause much larger issues, especially when missing edge cases. 

I would argue pretty strongly against going that route.

jwt/api.py
michaeldavis-wf: Looking at the ruby library that the original issue was based on, it looks like they use the `verify` kwarg instead of `verify_signature` in the options dictionary at all.  

I can see taking it out of the dictionary completely to be the same as the ruby library or doing something like this:

```python
merged_options = self._merge_options(override_options=options)
if verify or merged_options.get('verify_signature'):
    self._verify_signature(payload, signing_input, header, signature,
                                       key, algorithms)

self._validate_claims(payload, options=merged_options, **kwargs)
```

This would avoid the `verify` kwarg from overriding the other options and require that the developer double opt-in to not verifying the signature.

jwt/api.py
mark-adams: I support keeping these abbreviated so they are consistent with the claim names. This is similar to how the Ruby JWT library does things. I think we should keep them the way they are.

Regarding @johnlockwood-wf's comment about setting everything to False; I'm not sure we can avoid people being stupid with their usage. Defaulting to secure (and to the standard) is the best behavior here.

jwt/api.py
mark-adams: I think that giving people specific flags to turn on and off (like the PR does) is better than letting them pass in an entirely separate claims verifier.

If they really want to do something highly custom, they can always inherit from PyJWT.

jwt/api.py
```diff
@@ -25,6 +25,25 @@ def __init__(self, algorithms=None):
...
+            'verify_exp': True,
+            'verify_nbf': True,
+            'verify_iat': True,
+            'verify_aud': True,
+        }
+
+        try:
+            merged_options = dict()
+            for k, v in self.default_options.items():
+                merged_options[k] = options.get(k, v)
```
mark-adams: Couldn't this be replaced with:
```
merged_options = self.default_options.copy()
merged_options.update(options)
self.options = merged_options
```

?

jwt/api.py
```diff
@@ -233,6 +254,19 @@ def _validate_claims(self, payload, verify_expiration=True, leeway=0,
...
                 raise InvalidIssuerError('Invalid issuer')
 
+    def _merge_options(self, override_options=None):
+        if not override_options:
+            override_options = {}
+
+        try:
+            merged_options = dict()
+            for k, v in self.options.items():
+                merged_options[k] = override_options.get(k, v)
```
mark-adams: This looks very similar to my other comment about using `dict.update()` (see above)

This code looks very similar. You could probably break it out into a function and reuse it in both spots.

jwt/api.py
```diff
@@ -233,6 +254,19 @@ def _validate_claims(self, payload, verify_expiration=True, leeway=0,
             if payload.get('iss') != issuer:
                 raise InvalidIssuerError('Invalid issuer')
 
+    def _merge_options(self, override_options=None):
+        if not override_options:
+            override_options = {}
+
+        try:
+            merged_options = dict()
```
mark-adams: I'd like to see this changed to `{}`

jwt/api.py
```diff
@@ -25,6 +25,25 @@ def __init__(self, algorithms=None):
...
+        self.default_options = {
+            'verify_signature': True,
+            'verify_exp': True,
+            'verify_nbf': True,
+            'verify_iat': True,
+            'verify_aud': True,
+        }
+
+        try:
+            merged_options = dict()
```
mark-adams: I'd like to see this changed to {}

jwt/api.py
michaeldavis-wf: The slight difference is that the with your way, `self.options` can end up with keys that weren't originally in `default_dict`.  It isn't a problem as long as we don't ever assume `merged_options` to only contain the default keys.

mark-adams: That doesn't bother me since those extra keys won't affect the behavior of the library. If anything, we may introduce future extensibility so custom options may make sense.

I think the code is a cleaner without the loop too.

mark-adams:
Also, @michaeldavis-wf, make sure you add yourself to the AUTHORS file

jwt/api.py
johnlockwood-wf: How about adding common non-standard claims verification to the api? This way you have no single highly complex interface. Adding an options keyword argument that can take five different keys is the same as adding those keys to the method interface:

    decode(self, jwt, key='', verify=True, algorithms=None, verify_signature=False, verify_exp=False, verify_nbf=False, verify_iat=False, verify_aud=False, **kwargs)

The order of complexity is increased for each option.

As experts in the use cases, you will give less experienced users better information about how they should do things.

jwt/api.py
johnlockwood-wf: they should at least be labeled with expansions, or have a "constants" class that labels the keys, such as:

    class CLAIMS_KEYS(object):
        class VERIFY(object):
            EXPIRATION = 'verify_exp'
            ISSUED_AT = 'verify_iat'

jwt/api.py
johnlockwood-wf: People mess things up because they are hard and complicated and take a lot of knowledge to know if they are doing the right thing or not. If we can reduce the complication and give them an interface that better explains itself, they will be less likely to screw it up.

johnlockwood-wf: I think if the individual claims verification functionality was broken up,  it would only leave it to the more advanced user to compose them.

mark-adams: I agree with you that people screw things up because they are hard and complicated. However, I think that's why its simplest to present a secure set of defaults and make people aware of the fact that they can modify them at their own risk.

@michaeldavis-wf You might want to add to the README that changing the defaults is a behavior done at their own risk and may make their application less secure.

jwt/api.py
```diff
@@ -16,7 +16,7 @@
 
 
 class PyJWT(object):
-    def __init__(self, algorithms=None):
+    def __init__(self, algorithms=None, options=None):
```
johnlockwood-wf: what do you think about the keyword being, `verification_options`, so they are well categorized.

mark-adams: There's no reason why we couldn't use `options=` for other options later. I feel like `verification_options` would restrict what we could do with this argument unnecessarily.

In addition, all the current options start with `verify_` so I feel like it'd be a bit redundant to call it `verification_options`.

jwt/api.py
michaeldavis-wf: @mark-adams I updated the README with a note outlining that.

jwt/api.py
johnlockwood-wf: Grouping arguments into a category is better than having a lot of uncategorized arguments. If there are more unrelated  types of arguments in the future, they could be added to another group.

johnlockwood-wf: and to reduce redundancy, the verification_options dict could leave out the `verify_` prefix of the keys.

johnlockwood-wf: verify_claims = {'exp': False, 'aud': False}

johnlockwood-wf: As maintainers it is also easier to commit to a more restricted change to the interface.

jwt/api.py
mark-adams: > Adding an options keyword argument that can take five different keys is the same as adding those keys to the method interface

I disagree with you here entirely. The options object is much more readable than simple adding an argument to the method's arguments and that gives it an advantage over simply adding arguments.

I think maintaining some level of parity with ruby-jwt's API is also valuable here. (even though I'm not saying we should lock ourselves down to just matching ruby-jwt either; its just that I happen to agree with their design)

mark-adams: We can always iterate on this further. I think this design is a big improvement over what we have currently. I'm open to PRs suggesting further enhancements before we do a release.

The nice thing about code is that we can change it (until a release; then we have to consider API changes and how they will affect consumers)

mark-adams:
@michaeldavis-wf Last thing and I'm cool with merging this... can you add something to the CHANGELOG?

jwt/api.py
mark-adams: In the long-run, I'd rather add some sort of custom extensibility point (read: callable) that you could pass for additional validations on top of those provided by the library. 

I don't feel like we're at risk here of adding "too many unrelated arguments or options" even though I do understand your concern. We aren't going to add all sorts of crazy willy-nilly validations; our main focus is on parity with the validations with the JWT spec and I can't see it requiring us to add too many more options here.

mark-adams: I don't have a problem discussing more ideas in future PRs but I believe this is better than what we currently have for sure.

michaeldavis-wf:
@mark-adams @jpadilla Would it be possible to cut a release with this functionality in it, or possibly after https://github.com/jpadilla/pyjwt/pull/132 goes in?

jpadilla:
@mark-adams I'm fine with a new release whenever you want. You could create a PR with the version bump, etc and I'll go ahead and publish the release.

mark-adams:
Cool. I'll try to wrap things up tonight. I had a couple small changes I wanted to make

PR #132, opened 2015-04-09
Added support for RSASSA-PSS algorithms (PS256, PS384, PS512)

mark-adams:
Added support for RSASSA-PSS algorithms (PS256, PS384, PS512)

The JWT spec also mentions optional support for RSASSA-PSS using and MGF1. This is superior to PKCS #1 v1.5 (RS256, RS384, RS512) because it essentially uses a salt to randomize the padding instead of using a completely deterministic algorithm. No good attacks against PKCS #1 v1.5 padding currently exist, but PSS is recommended in new applications since it makes the signature even more random.

More info here: http://www.emc.com/emc-plus/rsa-labs/historical/raising-standard-rsa-signatures-rsa-pss.htm

Also added a test to cover our only uncovered line in `jwt.api`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 [Unreleased][unreleased]
 -------------------------------------------------------------------------
 ### Changed
+- Added flexible and complete verification options during decode #131
+- Added support for PS256, PS384, and PS512 algorithms. #132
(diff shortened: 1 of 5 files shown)


jpadilla:
Oh snap, this is great! As far as I know none of the other libraries do implement these.

mark-adams:
Most of the thanks goes to `cryptography` and OpenSSL. They did the hard parts.

PR #136, opened 2015-04-13
Changed to use Pytest as the test runner #111

### LINKED ISSUE #111: Use pytest

mark-adams:
Resolves #111 
* Installed Pytest
* Converted all unittest-related assertions to normal `assert` statements
* Removed all references to unittest from the project

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,3 +3,4 @@ omit =
     tests/*
     .tox/*
     setup.py
+    *.egg/*
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,5 @@
 [run]
-append = True
-source = jwt
+omit =
+    tests/*
+    .tox/*
+    setup.py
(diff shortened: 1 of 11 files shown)


jpadilla:
Great work with this! I'll look into [creating a pytest standalone script](https://pytest.org/latest/goodpractises.html#create-a-pytest-standalone-script) after this is in. It's what we use in [Django REST framework](https://github.com/tomchristie/django-rest-framework/blob/master/runtests.py).

mark-adams:
Any reason why we need the standalone Pytest script if `setup.py test` works? What benefit do we get?

jpadilla:
Not much benefit as is, other than considered "good practice", but I've grown fond of the one we use and having it run the tests as well as `flake8`.

mark-adams:
@jpadilla This is fixed-up and should be good to go now.

PR #139, opened 2015-04-16
README: Add example outputs as comments

Turbo87:


mark-adams:
Good point. I'd still like to see us use the doc-string convention though. I think it looks a little better

jpadilla:
@mark-adams you meant `>>>` instead of `>`?

```python
>>> import jwt
>>> jwt.encode({'some': 'payload'}, 'secret', algorithm='HS256')
'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzb21lIjoicGF5bG9hZCJ9.4twFt5NiznN84AWoo1d7KO1T_yoc0Z6XOpOVswacPZg'
```

I dig that better.

mark-adams:
Oh yea! That's what I meant.

Let's do that instead.

mark-adams:
Merging this and then I'll change the docstrings

PR #141, opened 2015-04-19
Added PyJWS

mark-adams:
So, eventually, I'd like to see us do more with JWS / JWE / JWK objects as well. (Probably as part of a 2.0 release).

At this point though, it makes sense to split up the JWS logic (signing and verification) from the JWT logic (JSON-encoding / -decoding of the payload and verification of the lcaims) into separate classes. That way it will be easier to make fixes to both branches (if we have bugs crop up) as we work on a 2.0.0 release branch over the next few months.

- Split `PyJWT` into `PyJWS` and `PyJWT` (inheriting from `PyJWS`)
- Split tests into JWS tests and JWT tests
- Split `PyJWT._validate_claims` into multiple `PyJWT._validate_<claim-name>` methods to simplify the code complexity a little bit


jpadilla:
@mark-adams oh yeah, this is a nice first step. I'm totally down for that. How about creating a new branch for v2 so we can instead merge all this work there in parallel to master?

mark-adams:
I think having a 2.0-develop branch would be a great idea! However, I think it might be a good idea to merge this PR into master, that way the JWS vs JWT logic will already be split out in master and any bugfixes that we need to make to the existing logic will be much less complicated to handle.

jpadilla:
@mark-adams gotcha, yea makes sense.

Issue #147, opened 2015-04-22
verify_expiration has been totally removed


cancan101:
It was here: https://github.com/jpadilla/pyjwt/blob/8f28b4a6124830fcea400668840e645bb91e38a2/jwt/api.py#L184
and now I see no mention of its removal / deprecation any docs, etc.

The only mention I see is:
>You can turn off expiration time verification with the verify_expiration argument.

mark-adams:
Yea, it looks like we missed that one. Using `{'verify_expiration': False}` for options should have the same effect. I submitted a PR to fix the README and CHANGELOG.

Thanks for raising the issue!

cancan101:
I do not think just adding a note to the docs is the right move.
At the very least the old parameter should be marked as deprecated and converted to the new parameter for at least one version.

mark-adams:
That's exactly what we did with the `verify` parameter to decode() since we would like to eliminate it at a later point. Our behavior with the deprecation of the verify parameter is very intentional and well documented (both in the CHANGELOG and using warnings)

The removal of `verify_expiration` is certainly something we wanted to do (for the same reasons we want to remove `verify`), but the actual removal of that parameter in the last release (instead of proper deprecation) slipped by us when we reviewed the PR.

At this point, I'm not sure I'm behind a knee-jerk restoration of the `verify_expiration` parameter since we already ripped the band-aid off. The only negative consequence is that the very few (probably) users who are using that parameter will run into verification issues if they are trying to decode an expired token. In other words, we're actually being a bit more secure. If things were the other way around and the removal of the parameter actually hurt security, I would not be opposed to bringing it back and making another release.

Unless others have a compelling argument, I'm for leaving it as is but documenting the change instead of making another release and having this weird "now its gone; now its back" thing going on.

@jpadilla, your thoughts?

cancan101:
The "few" users include @jpadilla's django-rest-framework-jwt which has not been updated.
Further one of the issues is that the that [`kwargs` is captured and then silently dropped](https://github.com/jpadilla/pyjwt/blob/ddbaf0008eac6559da5ebd1f24f68ee99ccb34b0/jwt/api_jwt.py#L76) which is annoying when debugging / dealing with mispelled params.

mark-adams:
Now I definitely agree with you there, I don't like our usage of **kwargs currently and I think that's something we should remove in a future release.

I'm sure @jpadilla would probably agree that django-rest-framework-jwt should be updated as well :-)

[...]

PR #149, opened 2015-04-22
Updated CHANGELOG and README to reflect removal of verify_expiration

mark-adams:
We forgot to document the fact we removed verify_expiration from decode()


mark-adams:
My suggestion for #147

PR #152, opened 2015-04-26
`verify_expiration` was removed too soon

jpadilla:
Fix for #147

- Merge with `verify_exp` option
- Add deprecation warning

@mark-adams thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 [Unreleased][unreleased]
 -------------------------------------------------------------------------
-### Changed
 ### Fixed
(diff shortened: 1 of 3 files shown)


mark-adams:
Looks good. Make sure you don't forget the CHANGELOG

jpadilla:
@mark-adams thoughts on what I should include? Is it just an item under "deprecation" or are we also adding(back) something?

mark-adams:
I would add a bugfix:
* Added back `verify_expiration=` argument to `jwt.decode()` that was erroneously removed in release X

And a deprecation warning:
* `verify_expiration=` argument to `jwt.decode()` is now deprecated and will be removed in a future version. Use the `option=` argument instead

Lothiraldan:
The verify_expiration parameter is still used at some places in the README

jpadilla:
@Lothiraldan good catch, will update now.

Issue #130, opened 2015-04-05
Removing bin/jwt?


jpadilla:
Do we want to keep this around or just remove it? How useful is it really? Are people using it?

mark-adams:
I didn't even know it was in there until three days ago. I could see how it might be useful but I feel like we might be able to fill the role by incorporating behavior into `python -m jwt` if people really use it.

mark-adams:
If we have tests for it, I think its not a bad idea to leave it around. I could see it being useful.

dresiu:
Maybe it can be useful, but it destroy my interpreter because if I want to import jwt and for example decode function, it says that "AttributeError: 'module' object has no attribute 'decode'" because te Scripts dir (on Windows) is before site-packages in paths. It could be renamed.

PR #153, opened 2015-04-29
Refactor CI Support

mindw:
* Added appveyor support (testing on Windows) https://ci.appveyor.com/project/mindw/pyjwt
* Refactored tox.ini to eliminate duplication using recent tox features.
* Refactor setup.py
* Convert jwt script to entrypoint - make it usable on Windows.
* Refactor setup.py to eliminate customizations.


mark-adams:
Looks like things aren't quite right yet...

mindw:
Yes it is expected. The jwt script was moved to reside inside the source tree but no tests for it exist. Only two configurations retained they original name so they reflect a 10% drop in coverage.

It was always there, undetected.

mark-adams:
Yea, we either need to write tests for that thing or get rid of it as discussed in #130 to get rid of this coverage issue.

mindw:
What about the contrib stuff? Isn't it redundant given the use of cryptography?

mark-adams:
The contrib stuff is basically the old pre-cryptography code that we reintroduced a few months back. Some users who were using PyJWT reported they were unable to use RSA on Google App Engine because it doesn't allow for compiling modules with C extensions. As a result, cryptography isn't an option so we brought the code back so they could fallback on PyCrypto if needed in GAE and similar scenarios.

mindw:
Ok. Please note that you need to setup and account on appveyor for its badge.

Any other comments for this pull request?

setup.py
```diff
@@ -3,66 +3,46 @@
...
+from setuptools import find_packages, setup
 
+__version__ = None
+with open('jwt/__init__.py') as fp:
+    _locals = {}
+    for line in fp:
+        if line.startswith('__version__'):
+            exec(line, None, _locals)
+            __version__ = _locals['__version__']
+            break
```
jpadilla: @mindw any specific reason for these changes?

setup.py
```diff
@@ -77,7 +57,17 @@ def get_packages(package):
         'Programming Language :: Python :: 3.4',
         'Topic :: Utilities',
     ],
+    # install_requires=['cryptography'],
     test_suite='tests',
-    tests_require=['pytest', 'pytest-cov'],
-    cmdclass={'test': PyTest},
```
mark-adams: Why are we removing PyTest from here? Won't this break our existing `setup.py test` functionality?

tests/contrib/test_algorithms.py
```diff
@@ -2,7 +2,7 @@
 
 import pytest
 
-from ..utils import ensure_bytes, ensure_unicode, key_path
+from tests.utils import ensure_bytes, ensure_unicode, key_path
```
mark-adams: The relative import makes more sense here to me unless you have a strong justification for the change.

tox.ini
```diff
@@ -1,56 +1,21 @@
 [tox]
-envlist = py26, py27, py27-contrib-crypto, py27-nocrypto, py32, py33, py34, py34-contrib-crypto, py34-nocrypto, pep8
+envlist = py{26,27,32,33,34}-crypto, py{27,34}-contrib_crypto, py{27,34}-nocrypto, flake8
```
mark-adams: Nice! I didn't realize it supported that type of expansion

setup.py
mark-adams: I agree, why are we manually reading this in instead of doing a simple `import` like we were previously?

.travis.yml
```diff
@@ -1,18 +1,19 @@
 language: python
-sudo: false
+sudo: true
```
mark-adams: Why do we need sudo here?

setup.py
mindw: The import happens before any of the project dependencies are installed and it will fail in case they are missing. A better variant og the above is to have a _version.py file in the project with the version and 'exec' it.

mindw: I'm looking into using setuptools_scm instead which will eliminate the version bookkeeping alltogher,

.travis.yml
mindw: We don't. I'll revert it.

tox.ini
mindw: Yep, makes tox.ini files really small.

tests/contrib/test_algorithms.py
mindw: I prefer them too. Something complained or didn't work - need to retest.

setup.py
mindw: Need to look into that - perhaps switching to the more explicit setup.py pytest is acceptable?

setup.py
jpadilla: Oh I see when it might actually fail. Let's do this instead.

```python
def get_version(package):
    """
    Return package version as listed in `__version__` in `init.py`.
    """
    init_py = open(os.path.join(package, '__init__.py')).read()
    return re.search("__version__ = ['\"]([^'\"]+)['\"]", init_py).group(1)

version = get_version('jwt')
```

That's what we use [here](https://github.com/tomchristie/django-rest-framework/blob/master/setup.py).

setup.py
jpadilla: Let's leave this as is for now.

mindw: Out of curiosity, where are we using setup.py test?

mark-adams: Its a setuptools convention: http://pythonhosted.org//setuptools/setuptools.html#test-build-package-and-run-a-unittest-suite

i.e. Anytime you see a setup.py, you should be able to run `setup.py test` and run the tests for the package.

setup.py
mindw: You can see the discussion here for recommendations:
http://stackoverflow.com/questions/2058802/how-can-i-get-the-version-defined-in-setup-py-setuptools-in-my-package

Please review my latest version which eliminates '__version__' extraction completely while providing automatic version management.

jpadilla:
@mindw I think next steps here should be simplifying this PR down to the changes for appveyor, travis, and tox. I'm mostly :-1: on setuptools-scm as I don't think that's currently an issue. I'd like a separate PR for actually removing `bin/jwt` so we can actually track that and document the change correctly. Thanks again for your work on this!

@mark-adams I'm I missing something else?

mindw:
@mark-adams My apologies for the delayed answer.

The issue at had is how to share the version number between the package itself and the metadata. Currently there are two possibilities:
1. put the version in the setup.py (metadata) - As implemented by this pull request.
2. put the version in the package.

The reasons why options 1 is better are:
1. A single place for the version - eliminate duplication and reduces chances for errors (version = latest tag).
2. Provides a distinct PEP440 compliant version for every unofficial build.
4. Eliminate redundant code from setup.py (DRY - don't repeat yourself).
5. Shorten release procedure (tag, sdist, upload).

The perceived negatives are the the setup_requires dependencies. Which is irrelevant when providing wheels.
 
@jpadilla advocates 2. Which does not have the above additional advantages but is a solution to the problem. 
In addition, The recommended implementation differs from the one suggested by @jpadilla. The recommended way is to place a file inside the package (_version.py) containing the version into the package and place the following into setup.py:
```python
_locals = {}
with open('jwt/_version.py') as fp:
    exec(fp.read(), None, _locals)
__version__ = _locals['__version__']
```
Simple and compatible with any [pep 440](https://www.python.org/dev/peps/pep-0440/). 

A good discussion can be found [here](http://stackoverflow.com/questions/2058802/how-can-i-get-the-version-defined-in-setup-py-setuptools-in-my-package).

mindw:
@jpadilla It is unlikely I can spare the time and effort to split this small pull into even smaller parts. If it will help I can do some minor fixups.
Should the decision would be for options 2 I'll do the extra work for it.

It was never my intention for this to consume so much time.

mark-adams:
No worries. I pulled the approved commits into a separate PR so we can get them merged in. (#157). Thanks for contributing!

Feel free to submit additional PRs (for #130 or anything else)!

PR #154, opened 2015-05-05
Made options a mandatory argument to PyJWT._validate_claims

### LINKED ISSUE #148: Default value for options in _validate_claims is silly
There is no reason to specify a default value for `options` in the `_validate_claims` method. A value of `None` will just result in an error.

https://github.com/jpadilla/pyjwt/blob/ddbaf0008eac6559da5ebd1f24f68ee99ccb34b0/jwt/api_jwt.py#L76

mark-adams:
Made options a mandatory argument to PyJWT._validate_claims and reordered the arguments to reflect the change. Resolves #148.

### DIFF
--- a/jwt/api_jwt.py
+++ b/jwt/api_jwt.py
@@ -69,12 +69,12 @@ class PyJWT(PyJWS):
 
         if verify:
             merged_options = merge_dict(self.options, options)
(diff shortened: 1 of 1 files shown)


jpadilla:
LGTM if you want to include it on v1.2.0

PR #156, opened 2015-05-08
Added get_unverified_header method

### LINKED ISSUE #155: Allow for the retrieval of unverified header values
@skion commented on #70 that it might be useful to have a way to parse out the unverified header parameters in the library prior to validating the entire JWT so that, for instance, a key could be looked up using the kid (until we get JWK support completed in 2.0)

We probably should create a jwt.utils function called `get_unverified_header()` or something like that.

mark-adams:
Added get_unverified_header method so that unverified headers can be retrieved by the application. Closes #155

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 [Unreleased][unreleased]
 -------------------------------------------------------------------------
+### Added
+- Added a new `jwt.get_unverified_header()` to parse and return the header portion of a token prior to signature verification.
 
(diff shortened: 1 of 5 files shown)


mark-adams:
Good point @skion. I added it to the global API.

jpadilla:
:+1: LGTM. Might be a good idea to always update the changelog for PR's and changes like this.

mark-adams:
Yep, I totally forgot. Just updated

jpadilla:
@mark-adams :+1:  thanks!

skion:
Thanks guys!

PR #160, opened 2015-05-19
Added test vectors from the IETF JOSE Cookbook for HMAC, RSA, and EC.

mark-adams:
The [JOSE Cookbook](https://www.ietf.org/id/draft-ietf-jose-cookbook-08.txt) has a very good set of test keys / signatures that we can use to verify our algorithms are functioning properly. Considering the recent discovery of our EC implementation being improper, I wanted to add these known good vectors to our test suite to verify everything is working properly prior to merging of #158.

It should be noted that there is a test in here for RSA-PSS that I'm not 100% I've written properly. I went ahead and left the test with a `@skipIf` decorator so that I can go in and fix it tomorrow once I know more about RSA-PSS.

Once this is merged, the EC test should show up as failing and we can use that test to verify #158 has fixed the issue.

### DIFF
--- a/tests/keys/__init__.py
+++ b/tests/keys/__init__.py
@@ -1,0 +1,72 @@
+import json
+import os
+
+from jwt.utils import base64url_decode
+
+from tests.utils import ensure_bytes, int_from_bytes
+
+BASE_PATH = os.path.dirname(os.path.abspath(__file__))
+
+
+def decode_value(val):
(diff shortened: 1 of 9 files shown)


mark-adams:
I'm merging this even though it is failing so the test can verify the #158 fix.

jpadilla:
@mark-adams this was a great idea, thanks for getting it done!

PR #163, opened 2015-05-20
RSA-PSS should use the proper salt value when creating and verifying signatures

mark-adams:
When we added the [RFC 7520](https://tools.ietf.org/html/rfc7520) test vectors to our project, the RSA-PSS algorithm test would not pass.

After further review, it turns out that I improperly implemented this originally. JWA [clearly specifies](https://tools.ietf.org/html/draft-ietf-jose-json-web-algorithms-31#section-3.5) how the salt length should be set:

> The size of the salt value is the same size as the hash function output.

This PR corrects that issue and removes some redundant tests from our original test suite for algorithms since the RFC 7520 tests accomplish the same thing.


jpadilla:
@mark-adams would you look at that! Great idea using those test vectors from the RFC. Merge this whenever you like.

PR #166, opened 2015-06-03
Added new options for requiring exp, iat, and nbf claims.

mark-adams:
Added the require_exp, require_iat, and require_nbf claims so that applications can indicate that they expect these claims to be present. If they are not present. These options are set to false by default.

A new exception type, MissingRequiredClaimError, is introduced. The `claim` property indicates which claim is missing. When `audience=` or `issuer=` is present on `decode()`, MissingRequiredClaimError is now raised instead of the previous exceptions.
 
Thanks to @dbaxa from @Atlassian for the suggestion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Exclude Python cache files from PyPI releases.
 
+### Added
+- Added new options to require certain claims
+  (require_nbf, require_iat, require_exp) and raise `MissingRequiredClaimError`
+  if they are not present.
+- If `audience=` or `issuer=` is specified but the claim is not present,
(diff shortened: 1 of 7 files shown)


jpadilla:
@mark-adams I like this. Only thing I see missing might be adding to docs.

Interested in knowing what would a scenario look like after catching and handling the `MissingRequiredClaimError` exception?

mark-adams:
Updated the docs. What do you think?

> ...what would a scenario look like after catching and handling the MissingRequiredClaimError exception?

I'm assuming that it would be very similar to most of our failure scenarios where the token would simply be rejected by the application. Having another exception subclass is mostly just to provide additional information and make it easier for us to raise the exception (by passing in the claim name and autogenerating the appropriate message)

jpadilla:
@mark-adams awesome, looks good to me. Merge whenever you like!

PR #174, opened 2015-07-21
Fail on encode and decode of bad JWS header values

gabrielg:
The [JWS spec states that if `kid` is present then it **MUST** be a string](https://tools.ietf.org/html/draft-ietf-jose-json-web-signature-41#section-4.1.4). Currently, the library allows silent creation of invalid JWS (and thus, JWT), as it allows any type for `kid`.

Allowing arbitrary types is both against the spec, and has some potential security ramifications that I've seen crop up. Since header values *can* be used pre-verification (and in the case of `kid`, might be required to verify), some more strictness around them and ensuring they meet the spec makes sense.

### DIFF
--- a/jwt/api_jws.py
+++ b/jwt/api_jws.py
@@ -6,7 +6,7 @@ from collections import Mapping
 
 from .algorithms import Algorithm, get_default_algorithms  # NOQA
 from .compat import string_types, text_type
-from .exceptions import DecodeError, InvalidAlgorithmError
+from .exceptions import DecodeError, InvalidAlgorithmError, InvalidTokenError
 from .utils import base64url_decode, base64url_encode, merge_dict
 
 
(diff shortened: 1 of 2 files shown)


jpadilla:
@gabrielg good catch. @mark-adams this looks good to me.

jwt/api_jws.py
```diff
@@ -180,6 +184,13 @@ def _verify_signature(self, payload, signing_input, header, signature,
         except KeyError:
             raise InvalidAlgorithmError('Algorithm not supported')
 
+    def _validate_headers(self, headers):
+        if 'kid' in headers:
+            self._validate_kid(headers['kid'])
+
+    def _validate_kid(self, kid):
+        if not isinstance(kid, string_types):
+            raise TypeError('Key ID header parameter must be a string')
```
mark-adams: I think this should probably use `InvalidTokenError` similar to other validation errors.

mark-adams:
:+1: Once you changeover that exception to be a `InvalidTokenError`, I'm ready to merge this.

The only reason we didn't have similar validation in here for that already was that we are preparing full JWK support as part of 2.0, but this seems like something reasonable to add now.

Thanks for contributing!

gabrielg:
Cool, changed `TypeError` over to `InvalidTokenError`. Thank you!

jpadilla:
@mark-adams should we release this already?

PR #178, opened 2015-08-04
If non-string passed as JWT, raise DecodeError

ariddell:
If you pass an integer to jwt.decode, you probably should get a DecodeError. Currently you get an attribute error here:

``AttributeError: 'int' object has no attribute 'rsplit'``


mark-adams:
Two bits of feedback:
* First, would you mind adding a test for this fix?
* Second, I think that catching `AttributeError` and raising `DecodeError` error with the same message is kind of confusing to a consumer. We should probably raise `DecodeError` with a different error message that more clearly tells the user what they did wrong

svartalf:
I think ValueError should be raised instead. It looks more pythonic for me.

mark-adams:
This was fixed in #184

PR #182, opened 2015-10-09
Fixes an issue where the newer versions of Pytest can't detect deprecation warnings

mark-adams:
Current versions of Pytest can't detect deprecation warnings.

This was fixed in https://github.com/pytest-dev/pytest/pull/1081 but is not released yet.


jpadilla:
Yar! I think we should also pin everything else including cryptography.

PR #184, opened 2015-10-22
Fixed #183 AttributeError: 'NoneType' object has no attribute 'rsplit'

### LINKED ISSUE #183: jwt.decode(None) causeAttributeError: 'NoneType' object has no attribute 'rsplit'
```python
>>> import jwt
>>> jwt.decode(None)
  File "C:\Python27\lib\site-packages\jwt\api_jwt.py", line 61, in decode
    payload, signing_input, header, signature = self._load(jwt)
  File "C:\Python27\lib\site-packages\jwt\api_jws.py", line 135, in _load
    signing_input, crypto_segment = jwt.rsplit(b'.', 1)
AttributeError: 'NoneType' object has no attribute 'rsplit'
>>> 
```

It should raise jwt.InvalidTokenError.

maoaiz:
Fixed #183 AttributeError: 'NoneType' object has no attribute 'rsplit'

The issue also occurs when payload is int raising:
    AttributeError: 'int' object has no attribute 'rsplit'
Test for None and int payload added

### DIFF
--- a/jwt/api_jws.py
+++ b/jwt/api_jws.py
@@ -139,7 +139,7 @@ class PyJWS(object):
             signing_input, crypto_segment = jwt.rsplit(b'.', 1)
             header_segment, payload_segment = signing_input.split(b'.', 1)
         except AttributeError:
-            raise DecodeError('Invalid payload type. It should be a {} not {}'.format(
+            raise DecodeError('Invalid payload type. It should be a {0} not {1}'.format(
                 text_type, type(jwt)))
         except ValueError:
             raise DecodeError('Not enough segments')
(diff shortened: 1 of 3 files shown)


mark-adams:
Hi @maoaiz!

Thanks for the contribution!

I'm not sure I care for catching `AttributeError` and then raising the decode error since this isn't really future-proof against other non-type related errors that might be thrown in that block.

What do you think about adding a check just above that block to verify that the token is the proper binary type? Maybe something like

```
if not issubclass(type(jwt), binary_type):
   raise DecodeError("Invalid token type. Token must be a {0}".format(binary_type))
```

where binary_type is defined as bytes or str in `jwt.compat`. I think that might be a little more future-proof and accomplish the same goal.

What do you think?

maoaiz:
I think you're right! I'm going to work on it. :+1:

mark-adams:
Looks great! Thanks again for contributing! :smiley:

PR #186, opened 2015-10-23
Make documentation awesome and add support for "Read the Docs"

mark-adams:
Hey @jpadilla! I finally got frustrated enough with our docs to do something about it.

What do you think? https://pyjwt.readthedocs.org/en/add-docs/


mark-adams:
Also, I should say this (just so its on the record):

**I HATE restructured text.** and **I :heart: Commonmark / Markdown**

That being said... RST is the prevailing standard for a lot of the Python world and really does make sweet documentation when coupled with Read The Docs so ¯\_(ツ)_/¯

jpadilla:
@mark-adams awesome, thanks for putting this together. I also hate reStructuredText. We could also use mkdocs for our docs, they could be hosted in GitHub Pages or Read the Docs. On the other side, you already did this and our docs aren't in constant flux. Thoughts?

mark-adams:
I'd love to use something like mkdocs but I'm not sure about switcfhing until they hit 1.0 (which their site claims will be soon)

Regarding Read the Docs, if I understand their docs correcftly, Sphinx's supports a lot more features with reStructuredText than it does with Markdown currently.

I wouldn't have a problem doing this (since its ready-to-ship) and then converting to something like mkdocs after they've hit 1.0

jpadilla:
@mark-adams sounds like a plan, merge whenever you like. Great work!

mark-adams:
Merged! Also, I added you as a "maintainer" in Read The Docs so you will have admin access to that stuff too.

jpadilla:
@mark-adams sweet, we should also probably link to docs from the README.

mark-adams:
I did add the Read the Docs badge that links to the docs already. What  else were you thinking?

jpadilla:
@mark-adams oops my bad, missed that completely.

PR #187, opened 2015-10-24
Renamed jwt CLI script to jwt-cli to avoid issues with jwt script overlapping with the jwt package.

mark-adams:
So, instead of getting rid of the jwt CLI script as we discussed in #130, what do you think about just renaming it to jwt-cli so that it doesn't overlap with `jwt` anymore? That should pretty much fix any outstanding issues.

Thoughts @jpadilla?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ Change Log
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+[Unreleased][unreleased]
+-------------------------------------------------------------------------
+### Changed
+- Renamed commandline script `jwt` to `jwt-cli` to avoid issues with the script clobbering the `jwt` module in some circumstances.
+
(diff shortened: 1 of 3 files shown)


jpadilla:
@mark-adams would this be a breaking change? People would be running `$ jwt-cli` now?

mark-adams:
Great point! Yes, this would be a breaking change. People would run `jwt-cli` instead of `jwt`.

I just updated the CHANGELOG to reflect this as well.

mark-adams:
@jpadilla I wanted to clarify one more time before I merged this... are you ok with this rename?

jpadilla:
TBH I don't really like adding `cli` to it. How about just calling it `pyjwt`?

mark-adams:
@jpadilla Done! Renamed jwt-cli -> pyjwt

jpadilla:
@mark-adams sweet, thanks! Should we put out a new release?

PR #188, opened 2015-10-26
Fixed example usage

sampwing:
The example contained incorrect syntax, simply swapped the comma for the existing period.


jpadilla:
Woops, thanks @sampwing!

PR #189, opened 2015-11-11
Fix imports in installation docs code example

mattijohn:


jpadilla:
Lol good catch, thanks!

jpadilla:
@mark-adams i'm not really sure how do we go about updating docs on RTD

Issue #190, opened 2015-11-17
Why validate that 'iat' is not in the future?


gobengo:
In https://github.com/jpadilla/pyjwt/blob/c5ee34e86bc42bef60ef6e701df569c2c86a5d5d/jwt/api_jwt.py#L129:

``` python
if iat > (now + leeway):
    raise InvalidIssuedAtError('Issued At claim (iat) cannot be in'
                               ' the future.')
```

I just debugged an issue in prod where `jwt.decode()` failed because of this. Mostly because the other party's jwt lib added 'iat' a few seconds or minutes ahead of our clock time ('clock skew' as mentioned in JWT specs).

I can't find any place in the specs that says that a JWT should be invalid if 'iat' is in the future. It seems like it's just there to be informative. I can use 'nbf' if I want to specify a ["time before which the token MUST NOT be accepted for processing"](https://tools.ietf.org/html/draft-ietf-oauth-jwt-bearer-12#section-3)

I consulted
- https://tools.ietf.org/html/rfc7519
- https://tools.ietf.org/html/draft-ietf-oauth-jwt-bearer-12

So either
1. I'm wrong and there is a JWT spec that says this is important to check. I want to know this, because if it's out there, I shouldn't just catch these errors from PyJWT and `pass`. Regardless of whether @jpadilla wants to remove that `raise` in his lib.
2. PyJWT is checking that unnecessarily, and we should remove it to be more compliant

soldnermike:
+1

mark-adams:
RFC 7519 says:
> The "iat" (issued at) claim identifies the time at which the JWT was issued.

In all circumstances except clock skew, it doesn't seem like it makes any sense for a JWT to be valid if, by its own claim, it hasn't been created yet. That's why we included this validation. JWT doesn't say you have to have this validation to be compliant, but it seems logical and enforces integrity in the data.

I'm against removing this validation, especially when you can already easily disable it by using ```jwt.decode(token, secret, options={'verify_iat': False})``` or account for the clock skew by using the leeway parameter: ```jwt.decode(token, secret, leeway=10)```

gobengo:
In all circumstances except clock skew

Clock skew is mentioned as twice in that RFC because it's unavoidable in federated distributed systems. It's not like that's not an important circumstance.

So should I fork pyjwt and make my another version that is only spec-compliant and doesn't include any extrastandard assertions that seem like they make sense but aren't in the RFC?

[...]

PR #191, opened 2015-11-18
Remove extrastandard assertion that iat is not from the future

gobengo:
Discussion on #190

### DIFF
--- a/jwt/api_jwt.py
+++ b/jwt/api_jwt.py
@@ -125,10 +125,6 @@ class PyJWT(PyJWS):
         except ValueError:
             raise DecodeError('Issued At claim (iat) must be an integer.')
 
-        if iat > (now + leeway):
-            raise InvalidIssuedAtError('Issued At claim (iat) cannot be in'
-                                       ' the future.')
-
     def _validate_nbf(self, payload, now, leeway):
         try:
             nbf = int(payload['nbf'])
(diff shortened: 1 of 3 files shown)


mark-adams:
So I think this is going too far. There's no reason to throw out a perfectly good feature of the library that some people may find desirable. In my case (and pretty much any case I can think of) a JWT generated in the future (which is impossible) is lying to me... why would I trust the rest of the claims?

I do understand your desire to bypass this functionality. I'd definitely support a PR that makes `verify_iat` default to `False`. I think that would alleviate your concerns while not removing a useful feature of the library. If you submit a PR that does that, I'm all for it.

gobengo:
@mark-adams @jpadilla I have amended the pull request to still leave iat future checking an option for those that want it.

However, this was not by changing the default value of the  'verify_iat' option to False, but by adding a new option.

This is because RFC7519 *does* specify how to verify iat (as NumericDate) with a MUST, and I don't think the default behavior should ignore that MUST. https://github.com/gobengo/pyjwt/commit/e8a243296677300f99f71699c00796dda2a34822#diff-00ea5fc2260636f8d87435daf8c3b4c7R138

But future checking, along with a specific exception that is a subclass of InvalidIssuedAtError, is still in there with an option. https://github.com/gobengo/pyjwt/commit/e8a243296677300f99f71699c00796dda2a34822#diff-6dc52fa39f2c4a084e5bde50b9537e4bR162

mark-adams:
This has been superseded by #252

PR #195, opened 2016-01-18
`get_unverified_headers` compat with python-jose

jpanganiban:


mark-adams:
I actually think the current name makes sense here without changes. RFC 7515 refers to the collection of header parameters collectively as the JWS / JWT "header" and each individual key as the "header paramter".

PR #199, opened 2016-03-02
Add password arg to encode and implement in EC and RSA_PSS algorithms:

leighklotz:
add optional password keyword arg to api_jws.PyJWS.encode
implement password argument to algorithms.RSAAlgorithm.prepare_key and algorithms.ECAlgorithm.prepare_key in calls to load_pem_private_key
add optional password keyword arg to algorithms._.prepare_key, and contrib/algorithms._.prepare_key, even for unimplemented ones
add to test_algorithms: test_rsa_pss_with_password_sign_then_verify_should_return_true, test_ec_with_password_sign_then_verify_should_return_true, and test_ec_sign_then_verify_should_return_true (base case for EC, which was missing)
add to test_api_jws: test_encode_decode_with_rsa_password_sha256, test_encode_decode_with_ecdsa_password_sha256
there are no encode tests in test_api_jwt.py so I didn't add ones for encrypted passwords


leighklotz:
fixed python 3.4 issues found in PR builder.

SaintDako:
+1

rogoman:
+1

jwt/algorithms.py
```diff
@@ -160,11 +160,15 @@ def prepare_key(self, key):
                 if isinstance(key, text_type):
                     key = key.encode('utf-8')
 
+                if isinstance(password, string_types):
```
jpadilla: @leighklotz shouldn't this just be:

```python
if isinstance(password, text_type):
    password = password.encode('utf-8')
```

jpadilla:
@mark-adams thoughts or feedback on this?

joeemprayil:
This will be great to have.

mark-adams:
@jpadilla I think I tend to prefer leaving this sort of thing out of the library. I think the library should handle the simple base case (a decrypted string of the public / private key) and anything more should be handled by the application code taking care of providing a `cryptography` `PublicKey` or `PrivateKey` object through whatever decoding or decrypting they need to do. I think adding more arguments to our existing methods clutter the API unnecessarily when you can already do this via alternate means.

What do you think?

jpadilla:
@mark-adams yup, that sounds right. We could probably add some examples to our docs around this.

mark-adams:
I've created #286 to handle updating the documentation to better include examples of how passphrases can already be used with the existing API.

PR #202, opened 2016-05-06
Add JWK support for HMAC and RSA keys

mark-adams:
This adds support for JWK serialization to HMAC and RSA keys.

Looking something like this:

```python
     public_key = RSAAlgorithm.from_jwk('{"kty": "RSA", "e": "AQAB", ...')
```

and also 

```python
     jwk = RSAAlgorithm.to_jwk(public_key)
     print(jwk)

     > '{"kty": "RSA", "e": "AQAB", ...'
```
```

### DIFF
--- a/jwt/algorithms.py
+++ b/jwt/algorithms.py
@@ -1,9 +1,15 @@
 import hashlib
 import hmac
+import json
 
-from .compat import binary_type, constant_time_compare, is_string_type
+
+from .compat import constant_time_compare, string_types
 from .exceptions import InvalidKeyError
-from .utils import der_to_raw_signature, raw_to_der_signature
+from .utils import (
+    base64url_decode, base64url_encode, der_to_raw_signature,
+    force_bytes, force_unicode, from_base64url_uint, raw_to_der_signature,
+    to_base64url_uint
+)
 
 try:
     from cryptography.hazmat.primitives import hashes
@@ -11,7 +17,8 @@ try:
         load_pem_private_key, load_pem_public_key, load_ssh_public_key
     )
     from cryptography.hazmat.primitives.asymmetric.rsa import (
-        RSAPrivateKey, RSAPublicKey
+        RSAPrivateKey, RSAPublicKey, RSAPrivateNumbers, RSAPublicNumbers,
+        rsa_recover_prime_factors, rsa_crt_dmp1, rsa_crt_dmq1, rsa_crt_iqmp
     )
     from cryptography.hazmat.primitives.asymmetric.ec import (
         EllipticCurvePrivateKey, EllipticCurvePublicKey
@@ -77,6 +84,20 @@ class Algorithm(object):
         """
         raise NotImplementedError
 
+    @staticmethod
+    def to_jwk(key_obj):
+        """
+        Serializes a given RSA key into a JWK
+        """
+        raise NotImplementedError
+
+    @staticmethod
+    def from_jwk(jwk):
+        """
+        Deserializes a given RSA key from JWK back into a PublicKey or PrivateKey object
+        """
+        raise NotImplementedError
+
 
 class NoneAlgorithm(Algorithm):
     """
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Comment by jpadilla** [comment]
@mark-adams let me know if there's anything I can help with.

**Comment by mark-adams** [comment]
@jpadilla Mind taking a look at this?

File: jwt/algorithms.py [hunk 1]
```diff
@@ -172,6 +204,104 @@ def prepare_key(self, key):
...
+
+            elif getattr(key_obj, 'verifier', None):
+                # Public key
+                numbers = key_obj.public_numbers()
+
+                obj = {
+                    'kty': 'RSA',
+                    'use': 'sig',
+                    'key_ops': ['verify'],
+                    'n': force_unicode(to_base64url_uint(numbers.n)),
```
clintonb: I cannot find specific documentation on encoding RSA keys, but the examples I have seen include modulus (n) and public exponent (e) values in both the private and public keys.

jwt/algorithms.py
```diff
@@ -172,6 +204,104 @@ def prepare_key(self, key):
...
+                    'qi': force_unicode(to_base64url_uint(numbers.iqmp))
+                }
+
+            elif getattr(key_obj, 'verifier', None):
+                # Public key
+                numbers = key_obj.public_numbers()
+
+                obj = {
+                    'kty': 'RSA',
+                    'use': 'sig',
```
clintonb: From https://www.rfc-editor.org/rfc/rfc7517.txt:
> The "use" and "key_ops" JWK members SHOULD NOT be used together;
   however, if both are used, the information they convey MUST be
   consistent.  Applications should specify which of these members they
   use, if either is to be used by the application.

This is valid; but, may be a potential source of error in the future.

jwt/algorithms.py
```diff
@@ -172,6 +204,104 @@ def prepare_key(self, key):
...
+        def from_jwk(jwk):
+            try:
+                obj = json.loads(jwk)
+            except ValueError:
+                raise InvalidKeyError('Key is not valid JSON')
+
+            if obj.get('kty') != 'RSA':
+                raise InvalidKeyError('Not an RSA key')
+
+            if 'd' in obj and 'e' in obj and 'n' in obj:
```
clintonb: Neither `'e'` nor `'n'` is included the JWK created above.

jwt/compat.py
```diff
@@ -5,7 +5,7 @@
 # flake8: noqa
 import sys
 import hmac
-
+import struct
```
clintonb: Minor: alphabetize these imports.

tests/test_algorithms.py
```diff
@@ -135,6 +170,162 @@ def test_rsa_verify_should_return_false_if_signature_invalid(self):
...
+            algo.to_jwk({'not': 'a valid key'})
+
+    @pytest.mark.skipif(not has_crypto, reason='Not supported without cryptography library')
+    def test_rsa_from_jwk_raises_exception_on_invalid_key(self):
+        algo = RSAAlgorithm(RSAAlgorithm.SHA256)
+
+        with open(key_path('jwk_hmac.json'), 'r') as keyfile:
+            with pytest.raises(InvalidKeyError):
+                algo.from_jwk(keyfile.read())
+
```
clintonb: Add a basic test to ensure `to_jwk` produces a JWK that can actually be passed to `from_jwk`:

```python
priv_key = ...
assert algo.from_jwk(algo.to_jwk(priv_key)) == priv_key
```

clintonb:
This looks good to go once the comments are addressed. Thanks, @mark-adams!

jwt/algorithms.py
mark-adams: Great point. I definitely missed that. I'll fix it.

tests/test_algorithms.py
mark-adams: Definitely would have caught the earlier mistake. I'll add it.

jwt/algorithms.py
mark-adams: Great point! I'll keep `key_ops` and ditch `use`

jwt/algorithms.py
mark-adams: You are correct. That was an oversight on my part. I missed the text that says:

> In addition to the members used to represent RSA public keys, the
> following members are used to represent RSA private keys.

mark-adams:
@jpadilla Does this look good to you?

jpadilla:
@mark-adams yup, all yours!

PR #203, opened 2016-05-06
Add Python 3.5 to CI builds

mark-adams:
Just noticed we don't have Python 3.5 set up for our CI builds. This adds Python 3.5 and changes some of our single version builds to use 3.5 instead of 3.4.


jpadilla:
@mark-adams awesome, thanks!

PR #213, opened 2016-08-04
Fix a bug where a PEM private key as bytes raises a TypeError

mark-adams:
On Python 3.4, if a private key is passed into `algo.prepare_key()` as a `bytes` object, it will raise a `TypeError`

The following test case demonstrates the issue.
```
def test_rsa_should_accept_pem_private_key_bytes(self):
    algo = RSAAlgorithm(RSAAlgorithm.SHA256)

    with open(key_path('testkey_rsa'), 'rb') as pem_key:
        algo.prepare_key(pem_key.read())
```

cc: @jpadilla

### DIFF
--- a/jwt/algorithms.py
+++ b/jwt/algorithms.py
@@ -1,7 +1,7 @@
 import hashlib
 import hmac
 
-from .compat import constant_time_compare, string_types, text_type
+from .compat import binary_type, constant_time_compare, is_string_type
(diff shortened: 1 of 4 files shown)


mark-adams:
@jpadilla Mind taking a look at this just to make sure I'm not doing anything crazy? :smile:

jpadilla:
@mark-adams ace! lgtm

PR #216, opened 2016-08-13
Added possibility to pass a passphrase to a encrypted RSA private key

desertkun:
I've noticed there is no way to encode a jwt token using encrypted RSA private key.

So I've added a possibility to do so via `**kwargs`:

```
access_token = jwt.encode(data, private_key, "RS256", password="private-key-passphrase")
```

### DIFF
--- a/jwt/api_jws.py
+++ b/jwt/api_jws.py
@@ -66,7 +66,7 @@ class PyJWS(object):
         return list(self._valid_algs)
 
     def encode(self, payload, key, algorithm='HS256', headers=None,
-               json_encoder=None):
+               json_encoder=None, **kwargs):
         segments = []
 
         if algorithm is None:
@@ -95,7 +95,7 @@ class PyJWS(object):
         signing_input = b'.'.join(segments)
         try:
             alg_obj = self._algorithms[algorithm]
-            key = alg_obj.prepare_key(key)
+            key = alg_obj.prepare_key(key, **kwargs)
             signature = alg_obj.sign(signing_input, key)
 
         except KeyError:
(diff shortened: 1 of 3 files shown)


jwt/api_jws.py
```diff
@@ -95,7 +95,7 @@ def encode(self, payload, key, algorithm='HS256', headers=None,
         signing_input = b'.'.join(segments)
         try:
             alg_obj = self._algorithms[algorithm]
-            key = alg_obj.prepare_key(key)
+            key = alg_obj.prepare_key(key, **kwargs)
```
mark-adams: So I'm not sure how I feel about passing `**kwargs` in here. We're basically saying that extra arguments to `encode()` are always destined for `prepare_key()`. What if we had additional arguments that we wanted to be passed to `sign()`?

desertkun: Well, there were no `kwargs` at all until this point. It could be considered to have one argument `password` instead. Don't have much time for now.

hoylen:
+1 for supporting passphrase protected RSA private keys.

Avoid the issue with `kwargs`, and simply define the _encode_ functions to take a named parameter with a default value: `passphrase=None`.

mark-adams:
Instead of cluttering up the main API by adding a `passphrase` arg, what do you think about simply adding a utility or helper function that would take a key string / bytes and a passphrase and return back a `PublicKey` or `PrivateKey`?

mark-adams:
I've created #286 to handle updating the documentation to better include examples of how passphrases can already be used with the existing API.

PR #219, opened 2016-09-22
Fix incorrectly named ECDSA algorithm

mike9005:
Per the [RFC on ECDSA in TLS](https://tools.ietf.org/html/rfc4492#appendix-A), the signing algorithm using the P-521 elliptic curve and SHA-512 is called ES521, instead of ES512.


mark-adams:
I agree with @jpadilla. At least until the next major version, this should still allow `ES512` and emit a warning if you use it.

PR #221, opened 2016-09-29
Convert README from Markdown to reStructuredText

adamchainz:
PyPI only supports reStructuredText, so this won't look broken there now.


jpadilla:
@adamchainz thanks, I don't mind this. Our docs use rst as well so what the heck. Mind resolving this conflicts so we can merge?

adamchainz:
fixed

jpadilla:
@adamchainz thanks!

PR #225, opened 2016-10-24
Add back 'ES512' for backward compatibility (for now)

mark-adams:
#219 renamed the incorrect `ES512` to `ES521` but we want to keep `ES512` for now in order to maintain backwards compatibility (for now)


mark-adams:
@jpadilla FYI

Issue #230, opened 2016-11-27
Is algorithm 'RS256' working


nv2908:
Is 'RS256' algorithm working? Giving 'Algorithm not supported' exception.

vimalloc:
Looks like you might need another dependency installed to use that RS256: https://github.com/jpadilla/pyjwt/issues/210

nv2908:
It worked. Thanks

jpadilla:
I've seen many people getting tripped by this. We should definitely improve error message here to suggest installing missing dependencies for those algorithms. Anyone want to tackle this?

vimalloc:
I could create a PR for this. Can likely have it done later today (if no
one else is actively working on it).

PR #231, opened 2016-11-28
Better error messages when missing cryptography package

vimalloc:
In reference to #230. If trying to use a cryptography algorithm without the cryptography package installed, you will get this error:

```
NotImplementedError: "cryptography" package must be installed to use this algorithm
```

instead of:

```
NotImplementedError: Algorithm not supported
```


vimalloc:
Whoops, looks like i based this on an older branch. I will update this.

tests/test_api_jws.py
```diff
@@ -303,6 +303,12 @@ def test_invalid_crypto_alg(self, jws, payload):
         with pytest.raises(NotImplementedError):
             jws.encode(payload, 'secret', algorithm='HS1024')
 
+    @pytest.mark.skipif(has_crypto, reason='Tests better errors if crypography not installed')
```
mark-adams: I think this would be better if it just said "Scenario requires that cryptography not be installed"

jwt/api_jws.py
```diff
@@ -101,7 +103,11 @@ def encode(self, payload, key, algorithm='HS256', headers=None,
             signature = alg_obj.sign(signing_input, key)
 
         except KeyError:
-            raise NotImplementedError('Algorithm not supported')
+            if not has_crypto and algorithm in get_crypto_algorithms():
+                raise NotImplementedError('"cryptography" package must be '
+                                          'installed to use this algorithm')
```
mark-adams: How about something like `Algorithm %s could not be found. Do you have cryptography installed?`

jwt/algorithms.py
```diff
@@ -44,22 +73,21 @@ def get_default_algorithms():
...
-            'PS384': RSAPSSAlgorithm(RSAPSSAlgorithm.SHA384),
-            'PS512': RSAPSSAlgorithm(RSAPSSAlgorithm.SHA512)
-        })
+        crypto_algorithms = _get_crypto_algorithms()
+        default_algorithms.update(crypto_algorithms)
 
     return default_algorithms
 
 
+def get_crypto_algorithms():
```
mark-adams: Instead of the `get_crypto_algorithms()` approach, I think it might be simpler to just use a set to define the cryptography-supplied algorithms.

Something like:

```
requires_cryptography = set([
    'RS256', 'RS384', 'RS512', 'ES256', 'ES384', 'ES521',
    'PS256', 'PS384', 'PS512'
])
```

Then instead of `if not has_crypto and algorithm in get_crypto_algorithms()`, you could simply check `if algorithm in requires_cryptography`.

jwt/__main__.py
```diff
@@ -131,5 +131,6 @@ def main():
     else:
         p.print_help()
 
+
```
mark-adams: Looks like you might have added an extra space here?

jwt/algorithms.py
```diff
@@ -171,6 +199,7 @@ def sign(self, msg, key):
     def verify(self, msg, key, sig):
         return constant_time_compare(sig, self.sign(msg, key))
 
+
```
mark-adams: Another extra space?

vimalloc: This (and the other extra space) was to fix flake8 errors reported by the tests. It seemed to be failing to pass without the changes (but maybe that was just a local configuration on my box)

vimalloc:
Thx for the feedback. I'll make any changes tomorrow and get them pushed to you. Cheers.

jwt/algorithms.py
vimalloc: er. Extra newline, not extra space.

mark-adams: It's possible our `flake8` in CI just started caring about these. If so, carry on. As long as flake8 is happy, I'm happy.

jwt/algorithms.py
mark-adams: Yea, I think that it requires a little extra work to update but it is a little simpler in the actual implementation.

mark-adams:
One last thing... do you think you could add a line to the CHANGELOG about this?

vimalloc:
Sure. Just a sec

mark-adams:
Thanks for contributing @vimalloc! :+1:

jpadilla:
@vimalloc @mark-adams thanks! 🎉

jpadilla:
@mark-adams there's nothing else pressing, should we release v1.4.3 with this?

Issue #206, opened 2016-06-22
TypeError in compat.py with Python 3.2.5


shamdan:
File "/usr/local/site-packages/jwt/api_jwt.py", line 64, in decode
    options, **kwargs)
    File "/usr/local/site-packages/jwt/api_jws.py", line 115, in decode
    key, algorithms)
    File "/usr/local/site-packages/jwt/api_jws.py", line 177, in _verify_signature
    if not alg_obj.verify(signing_input, key, signature):
    File "/usr/local/site-packages/jwt/algorithms.py", line 138, in verify
    return constant_time_compare(sig, self.sign(msg, key))
    File "/usr/local/site-packages/jwt/compat.py", line 50, in constant_time_compare
    result |= ord(x) ^ ord(y)
    TypeError: ord() expected string of length 1, but int found

This is what I get when using `decode`. I am using [pypy3 v2.4](http://doc.pypy.org/en/latest/release-pypy3-2.4.0.html)

jpadilla:
We're not testing against pypy, so I'm not sure what our support for it might be. A pull request adding those to Tox/Travis is welcome. We can see how to move forward from there.

shamdan:
I ended up upgrading pypy version which now supports Python 3.3. Thanks

efossier:
@jpadilla PyJWT doesn't support python 3.2 because hmac.compare_digest was added in 3.3, it therefore falls back to your python2 reimplementation of compare_digest which does an ord() on the characters before comparing them. This works in python2 but in python3 zip() on bytes objects returns integers giving the error above. 

I fixed this in my project by re-implementing hmac.compare_digest without the ord calls again and monkey patching that method onto hmac.compare_digest so PyJWT finds it.

I also created this PR to fix this in master if you are interested: https://github.com/jpadilla/pyjwt/pull/237

PR #237, opened 2017-02-21
Support Python 3.2

efossier:
Fixes https://github.com/jpadilla/pyjwt/issues/206

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ env:
   - TOXENV=flake8
   - TOXENV=py26-crypto
   - TOXENV=py27-crypto
+  - TOXENV=py32-crypto
   - TOXENV=py33-crypto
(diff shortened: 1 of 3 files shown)


jpadilla:
@mark-adams do you remember where we stand on supporting 3.2?

mark-adams:
@jpadilla I believe we stopped supporting 3.2 because it had too many 3.2-specific issues and it seems like many mainstream libraries were starting to drop support as well.

efossier:
@jpadilla @mark-adams Ah ok, didnt realize that was a deliberate decision. Feel free to close.

jpadilla:
@efossier thanks, closing for now.

leason:
For what it's worth I believe I just ran into the issue that caused this PR to be opened.  I'm running Python 3.6.3 and PyJWT 1.5.3.

PR #238, opened 2017-02-26
Change optparse for argparse.

froi:
- Change optparse for the more modern argparse.
- Refactored main() to fit new argparse code.
- Added two subcommands: encode and decode
- Added default functions for encode and decode subcommands
- Replaced implicit arguments with standard Python syntax in usage text.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ env:
   - TOXENV=py35-contrib_crypto
   - TOXENV=py36-contrib_crypto
   - TOXENV=py27-contrib_crypto
+  
 install:
   - pip install -U pip
   - pip install -U tox coveralls
--- a/jwt/__main__.py
+++ b/jwt/__main__.py
@@ -148,20 +148,16 @@ def build_argparser():
     return arg_parser
 
 
-def main(args):
+def main():
     arg_parser = build_argparser()
 
     try:
-        arguments = arg_parser.parse_args(args)
+        arguments = arg_parser.parse_args(sys.argv[1:])
 
-        return arguments.func(arguments)
+        output = arguments.func(arguments)
+
+        print(output)
     except Exception as e:
         print('There was an unforseen error: ', e)
         arg_parser.print_help()
         sys.exit(1)
-
-
-if __name__ == '__main__':
-    output = main(sys.argv[1:])
-
-    print(output)
--- a/jwt/__main__.py
+++ b/jwt/__main__.py
@@ -160,4 +160,3 @@ def main():
     except Exception as e:
         print('There was an unforseen error: ', e)
         arg_parser.print_help()
-        sys.exit(1)
--- a/jwt/__main__.py
+++ b/jwt/__main__.py
@@ -153,6 +153,7 @@ def build_argparser():
 
     return arg_parser
 
+
 def main(args):
     arg_parser = build_argparser()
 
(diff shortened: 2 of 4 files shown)


jpadilla:
@froi looks good to me, thanks! Coverage decreased because `encode_payload()` and `decode_payload()` add API surface which we didn't have before. We currently don't have any tests for the CLI tool.

froi:
Let me see if I can make that happen in a sane way. Do you want me to add that to this PR or another?

jpadilla:
@froi here's fine.

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: python
 python:
-  - "3.5"
+  - "3.6"
```
jpadilla: Is changing these required for this PR? There are some `InterpreterNotFound` errors in there, which might be fixed by #262.

tests/test_cli.py
```diff
@@ -0,0 +1,60 @@
+
```
jpadilla: This file has a couple of flake8 errors

```
./tests/test_cli.py:2:5: E271 multiple spaces after keyword
./tests/test_cli.py:3:1: F401 'time' imported but unused
./tests/test_cli.py:3:1: I201 Missing newline before sections or imports.
./tests/test_cli.py:7:1: I201 Missing newline before sections or imports.
```

https://travis-ci.org/jpadilla/pyjwt/jobs/232649651

tests/test_cli.py
```diff
@@ -0,0 +1,60 @@
...
+            decode_payload(args)
+
+        assert 'There was an error decoding the token' in str(excinfo.value)
+
+    @pytest.mark.parametrize('key,name,job', [
+        ('1234', 'Vader', 'Sith'),
+        ('4567', 'Anakin', 'Jedi'),
+    ])
+    def test_main_run(self, key, name, job):
+        args = [
```
jpadilla: Two errors here in 2.6 https://travis-ci.org/jpadilla/pyjwt/jobs/232649652

Changing it to the following should fix.

```python
args = [
    '--key', key,
    'encode',
    'name={0}'.format(name),
    'job={0}'.format(job)
]
```

froi: Fixed here - f6612a7

tests/test_cli.py
froi: Fixed here - f6612a7

jwt/__main__.py
```diff
@@ -2,135 +2,172 @@
...
+        arguments = arg_parser.parse_args(args)
+
+        return arguments.func(arguments)
+    except Exception as e:
+        print('There was an unforseen error: ', e)
+        arg_parser.print_help()
+        sys.exit(1)
 
 
 if __name__ == '__main__':
```
jpadilla: @froi Noticed another thing missing.

1. Run `python setup.py install`
2. Run `pyjwt`

It'll fail with something like this:

```
$ pyjwt
  File "/Users/jpadilla/.pyenv/versions/pyjwt/bin/pyjwt", line 9, in <module>
    load_entry_point('PyJWT==1.5.0', 'console_scripts', 'pyjwt')()
TypeError: main() takes exactly 1 argument (0 given)
```

Since we're using the `console_scripts` entry point on our `setup.py` I think we should be able to drop just drop the `if __name__ == '__main__': ...` and have the `main()` method optionally set args to `sys.argv[1:]`. Let me know if this makes sense.

froi: That'll work but it'll also make `main()` harder to test I think.

jpadilla:
@froi Thank you!

Issue #205, opened 2016-06-01
Allow a list of valid audiences to be configured


jobec:
A JWT token can contain multiple audiences.

But, it would be nice if you could also specify a list of valid audiences for `audience` in `jwt.decode()`.

Example:

``` Python
options = {
    'verify_signature': True,
    'verify_exp': True,
    'verify_nbf': True,
    'verify_iat': True,
    'verify_aud': True,
    'verify_iss': True,
    'require_exp': False,
    'require_iat': False,
    'require_nbf': False
}

payload = jwt.decode(
    jwt_value,
    key=self._public_key,
    verify=True,
    audience=["audience_1", "audience_2"],
    issuer="SomeIssuer",
    options=options,
)
```

mark-adams:
I think the reason this wasn't implemented originally was based primarily on the typical use case.

For instance, a normal JWT use case is for an identity provider (the issuer) to issue a token to a user that is valid for service A, service B, and service C. When service A receives a request containing the token, it doesn't care about whether or not service B or service C are in the list, only that it (service A) is in the list of audiences on the token.

For that reason, I think most use cases only require validating that a single audience is in the list.

AndrewJHart:
I actually agree with @jobec because in our instance - using JWT for oauth - we have 2 different audiences. One audience is a single page app that once authenticates with users service has a unique client_id & audience. But, if a service needs to communicate with another service - when it makes a request to the users service it will have a different audience/client_id. Therefore, since we can only set one audience - we're limited to either allowing only the single page app to make API requests to services or limited to allowing only services to make requests to other services. 

I would be happy to work on this if you would give it the ok - I don't think openid-connect and oauth flows were originally considered when designing the audience check for this library... 

Thanks,

codervinod:
@mark-adams We have a use case where identity provider issues tokens to different clients. We have a single API which authenticates via JWT and hence we need to absorb a list of audience. I will send a pull request regarding this feature later today. It would be helpful if you can merge it.

PR #246, opened 2017-03-15
Resolved issue #205, Allow a list of valid audiences to be configured

codervinod:
Added test cases to check if audience is provided a list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,6 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Add support for ECDSA public keys in RFC 4253 (OpenSSH) format [#244][244]
 - Renamed commandline script `jwt` to `jwt-cli` to avoid issues with the script clobbering the `jwt` module in some circumstances.
 - Better error messages when using an algorithm that requires the cryptography package, but it isn't available [#230][230]
-
-### Fixed
-- Allow a list of valid audiences to be configured [#205][205]
-
-[v1.4.3][1.4.3]
 -------------------------------------------------------------------------
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Better error messages when using an algorithm that requires the cryptography package, but it isn't available [#230][230]
 
 ### Fixed
+- Allow a list of valid audiences to be configured [#205][205]
+
+[v1.4.3][1.4.3]
+-------------------------------------------------------------------------
+
+### Fixed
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Better error messages when using an algorithm that requires the cryptography package, but it isn't available [#230][230]
 
 ### Fixed
+- Allow a list of valid audiences to be configured [#205][205]
+
+[v1.4.3][1.4.3]
```
mark-adams: No need to add this version section here. We do that whenever we make new releases.

jwt/api_jwt.py
```diff
@@ -87,8 +87,13 @@ def _validate_claims(self, payload, options, audience=None, issuer=None,
         if isinstance(leeway, timedelta):
             leeway = timedelta_total_seconds(leeway)
 
-        if not isinstance(audience, (string_types, type(None))):
-            raise TypeError('audience must be a string or None')
+        if not isinstance(audience, (string_types, type(None), list)):
+            if not isinstance(audience, list):
+                raise TypeError('audience must be a string or None')
```
mark-adams: What do you think about moving this all into `self._validate_aud()`?

That would probably simplify the implementation a little bit.

mark-adams: Actually, it looks like you basically already did that for the most part. Maybe we should just get rid of the redundant check here completely?

CHANGELOG.md
mark-adams: We typically add the new version section when we do a release. We can probably delete these lines for now and leave it up to the maintainers at release time.

codervinod: Done

jwt/api_jwt.py
codervinod: Done

codervinod:
Hi Mark Adams,

Thanks for review. I took care of your comments. Could you please review?

Thanks.
Vinod

jwt/api_jwt.py
```diff
@@ -161,8 +158,17 @@ def _validate_aud(self, payload, audience):
...
-            raise InvalidAudienceError('Invalid audience')
+
+        if isinstance(audience, string_types):
+            audience = [audience]
+
+        if not isinstance(audience, list):
+            raise InvalidAudienceError('Invalid audience format')
+
+        for aud in audience:
+            if aud in audience_claims:
```
mark-adams: Since we are now checking multiple values to see if they are in `audience_claims` do you think it might make more sense to make `audience_claims` a set?

jpadilla:
Closed by #306

PR #249, opened 2017-03-28
Update FAQ to clarify Python2/3 differences

dbinetti:
In response to #319


dbinetti:
sure, although most lines will be the same... not very DRY :-).

jpadilla:
@dbinetti as it is I feel like somebody on Python 3 would try copy pasting most of the code from the snippet, it'll fail, then they might go back and see the note on Python 3.

dbinetti:
I've actually never edited a PR before; is that all I need do?

jpadilla:
@dbinetti yep, that's it. Thanks!

PR #252, opened 2017-04-17
Remove rejection of future 'iat' claims

### LINKED ISSUE #190: Why validate that 'iat' is not in the future?

mark-adams:
This change resolves #190 by no longer rejecting future `iat` claims.

RFC 7519 does not require or even mention this type of validation so it seems best to leave this up to applications.

In addition, this PR changes the validation that rejects non-numeric `iat` values to raise `InvalidIssuedAtError` instead of `DecodeError`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Add support for ECDSA public keys in RFC 4253 (OpenSSH) format [#244][244]
 - Renamed commandline script `jwt` to `jwt-cli` to avoid issues with the script clobbering the `jwt` module in some circumstances.
 - Better error messages when using an algorithm that requires the cryptography package, but it isn't available [#230][230]
+- Tokens with future 'iat' values are no longer rejected [#190][190]
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


vergenzt:
FYI this was re-introduced by #794

PR #262, opened 2017-05-13
enable testing with python 3.6

purificant:
Enable running tests against python 3.6 in Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,12 @@
 language: python
 python:
-  - "3.5"
+  - "3.6"
+addons:
+  apt:
+    sources:
+      - deadsnakes
+    packages:
+      - python3.5 python3.5-dev
 sudo: false
 env:
   - TOXENV=flake8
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,6 +1,12 @@
 language: python
 python:
-  - "3.5"
+  - "3.6"
+addons:
+  apt:
+    sources:
+      - deadsnakes
+    packages:
+      - python3.5 python3.5-dev
```
jpadilla: Why is this `addons` section required?

purificant: Hi @jpadilla, thanks for taking a look at the PR.

This section installs python3.5 so that both 3.5 and 3.6 interpreters are available in the base test image during Travis run.
"python3.5" is needed to run tests, "python3.5-dev" is needed for crypto dependencies.

Without this section you get an error `ERROR: InterpreterNotFound: python3.5`

It appears to be an issue with travis base images as described here: https://github.com/travis-ci/travis-ci/issues/4794, the solution also comes from discussions in that thread, I've been iterating on it here: https://travis-ci.org/purificant/pyjwt/builds 

Once the issue is resolved upstream we can remove this section.

jpadilla: @purificant great, thanks for that!

PR #270, opened 2017-06-12
DeprecationWarnings: more helpful message

blueyed:
I've thought about adding `stacklevel=2`, but it is used internally / passed through, so that might not be that helpful.


jpadilla:
@blueyed good stuff, thanks!

PR #271, opened 2017-06-12
PyJWT.decode: move verify param into options

blueyed:
Followup to https://github.com/jpadilla/pyjwt/pull/270:

It seems that "verify" is only deprecated with `PyJWS.decode`, which
makes sense, since you would have to override a lot of options to skip
the verification in `PyJWT._validate_claims`.

This makes `PyJWT.decode` use the `verify_signature` option when calling
`PyJWT.decode`, and therefore also allows to use `stacklevel=2` there
then for the DeprecationWarning.

### DIFF
--- a/jwt/api_jws.py
+++ b/jwt/api_jws.py
@@ -127,7 +127,7 @@ class PyJWS(object):
         else:
             warnings.warn('The verify parameter is deprecated. '
                           'Please use verify_signature in options instead.',
-                          DeprecationWarning)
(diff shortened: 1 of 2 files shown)


blueyed:
Please let me know if I understood it correctly now, and if the fixup is better than the original in your opinion.
I'll squash it then.

jpadilla:
@blueyed looks good to me 👍

blueyed:
@jpadilla 
Thanks for the review, squashed.

PR #272, opened 2017-06-15
fix issuer bug

felixglow:
`encoded = jwt.encode({'name': 'jack', 'exp': datetime.datetime.utcnow(), 'iss': 'urn:foo'}, 'welcome', algorithm='HS256')`

it's ok
```
jwt.decode(encoded, 'welcome', algorithm='HS256', leeway=100, issuer='23432')
InvalidIssuerError: Invalid issuer
```
but:
```
jwt.decode(encoded, 'welcome', algorithm='HS256', leeway=100)
{u'exp': 1497512206, u'iss': u'urn:foo', u'name': u'jack'}
```

### DIFF
--- a/jwt/api_jwt.py
+++ b/jwt/api_jwt.py
@@ -166,10 +166,10 @@ class PyJWT(PyJWS):
             raise InvalidAudienceError('Invalid audience')
 
     def _validate_iss(self, payload, issuer):
-        if issuer is None:
+        if issuer is None and 'iss' not in payload:
             return
 
-        if 'iss' not in payload:
+        if issuer is not None and 'iss' not in payload:
             raise MissingRequiredClaimError('iss')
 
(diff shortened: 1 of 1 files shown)


mark-adams:
I'm not sure why this is a bug? If you don't pass `issuer=` to `decode()`, we're assuming that you don't care about the `iss`. The JWT spec does not specify any sort of mandatory validation logic on `iss` if it is present. The only reason we even have the `issuer=` arg is to save the consumer some code if they care to only accept tokens with a specific `iss` (a common case). I would expect a lot of JWT issuers would specify an `iss` but many clients might not care about who the `iss` actually is as long as it passes signature verification.

Maybe I'm missing it but this doesn't look like a bug.

mark-adams:
No response on this for a few months. Closing it out.

PR #273, opened 2017-06-15
Add sort_keys for json.dumps in jws/jwt encoding

james-jra:
PASSporT is a [draft-ietf-stir-passport-11](https://tools.ietf.org/html/draft-ietf-stir-passport-11) uses JWT.
[Section 9](https://tools.ietf.org/html/draft-ietf-stir-passport-11#section-9) of the above document mandates that the JSON objects in the payload and headers must be ordered by key.

I realise this isn't part of the JWT spec, but it's a really basic change.

Changes:

* add sort_json argument for jwt and jws encode() methods
    * simply passes through to existing sort_keys argument in json.dumps()
* add --sort-payload command line argument for encode command

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ develop-eggs/
 dist/
 downloads/
 eggs/
+.eggs/
 lib/
 lib64/
 parts/
--- a/jwt/__main__.py
+++ b/jwt/__main__.py
@@ -46,7 +46,8 @@ def encode_payload(args):
     token = encode(
         payload,
         key=args.key,
-        algorithm=args.algorithm
+        algorithm=args.algorithm,
+        sort_json=args.sort_payload
     )
 
     return token.decode('utf-8')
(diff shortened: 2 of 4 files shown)


mark-adams:
Here's an interesting question: what would be the downside to always sorting the keys? Since JWT doesn't specify any sort of ordering, I wouldn't see interoperability being an issue.

The biggest problem I could imagine is the length of time it takes to sort keys on large dicts but I'd like to see a benchmark of the difference in timings.

james-jra:
I ran [this](https://github.com/akrylysov/python-json-benchmark.git) benchmark (on a MacBook Pro late 2013 Intel i5 2.4GHz)
Each runs 100 times

`json.dumps` without sorting:

- 1 x large object: 0.614s
- 1000 x small objects: 1.045s

`json.dumps` with sorting:

- 1 x large object: 3.751s
- 1000 x small objects: 1.067s

mark-adams:
It seems like the existing API already can do what you want. 

Instead of adding another parameter to `encode()`, what if you just used the existing `json_encoder` and passed a JSON encoder that sorts the keys by default (i.e. subclass of `JSONEncoder` with `sort_keys` set to True)?

james-jra:
That's true. Thanks for the help!

PR #274, opened 2017-06-16
Update algorithms.rst

bhavin177:


mark-adams:
This seems extraneous. Why do you think this is necessary?

If this is needed, I think I'd rather see an abbreviated example in a code block outside the sentence instead of crammed into the text itself.

I'd expect most developers to be able to google PEM format is if they are unfamiliar. I'm not sure it needs to be expressly spelled out like that.

bhavin177:
Regarding need: The issue, as I experienced it, isn't with specificity of the `PEM` format, which can indeed be Googled - it is with confusing how to pass the key-file to PyJWT. I think the appropriate way is to show the key file being read into a variable which is then passed.

I do see your point around the current edit - seems a bit noisy. 

My initial intent was to add a code block for  `ECDSA` but then one would have to enforce consistency with all other code-blocks (for other Algorithms, etc.) - which I personally don't have enough experience to feel comfortable modifying. This, otherwise, would be ideal.

Less optimal but still a decent option - I can add a FAQ `How to generate Apple Developer Token`. Why you may ask? Specifically, Apple Music just opened up their API which requires a Dev Token. Their help docs. point to jwt.io, which in turn point to PyJWT - so few devs will be looking for this.

Interested in our thoughts.

bhavin177:
Specifically, here is what I was thinking:

```
keyfile = open ("AuthKey_xxxxxx.p8", 'r')
keystring = keyfile.read()
token = jwt.encode({"iss": "abc123"}, keystring, algorithm='ES256',headers={'kid': 'xyz456'})
```

rsussland:
Hello. the current algorithms.rst only contains examples of digital signatures and not encryption. Could you please provide a list of the encryption algorithms that are available as well? I'm interested in RSA-OAEP-256 :)

PR #277, opened 2017-06-22
Guard against PKCS1 PEM-encoded public keys

jpadilla:
Previous to this we were not correctly accounting for all PEM encoded public keys, like the PKCS1 PEM encoded format which is prefaced by `-----BEGIN RSA PUBLIC KEY-----`. This fix prevents symmetric/asymmetric key confusion attacks against users using the PKCS1 PEM encoded public keys, which would allow an attacker to craft JWTs from scratch.

We've also added a deprecation warning when using `decode()` and not specifying the `algorithms` param. This will be required in a future release.


paragonie-scott:
Should this also block ECDSA public keys? Or are those outright not supported?

Issue #279, opened 2017-06-27
Proposal for alternative to require_xxx options


psafont:
Hi,

I'd like to have a flexible way to verify some claims that are not dealt with in `api_jwt.py`'s `_validate_claims`.

I find that having an option like `'require': ['exp', 'iat', 'email']` would suit my needs while not taking away any functionality from the library.
I'd also like to know your opinion on how could I provide backwards-compatibility, I'd be implementing this enhancement, of course.

PR #280, opened 2017-06-28
Require tweak

psafont:
Goes ahead with the proposal in https://github.com/jpadilla/pyjwt/issues/279

Added a test and documentation on how to use the new parameter.
I took the liberty to deprecate the 'verify_xxx' options while still allowing them (as evidenced by the existing tests)

Feel free to critique any of the changes.

### DIFF
--- a/jwt/api_jwt.py
+++ b/jwt/api_jwt.py
@@ -26,10 +26,11 @@ except ImportError:
 
 class PyJWT(PyJWS):
     header_type = "JWT"
+    deprecated_requires = ["require_exp", "require_iat", "require_nbf"]
 
     @staticmethod
     def _get_default_options():
-        # type: () -> Dict[str, bool]
+        # type: () -> Dict[str, Union[bool, List[str]]]
         return {
             "verify_signature": True,
             "verify_exp": True,
@@ -37,9 +38,7 @@ class PyJWT(PyJWS):
             "verify_iat": True,
             "verify_aud": True,
             "verify_iss": True,
-            "require_exp": False,
-            "require_iat": False,
-            "require_nbf": False,
+            "require": [],
         }
 
     def encode(
@@ -128,6 +127,22 @@ class PyJWT(PyJWS):
                 DeprecationWarning,
             )
 
+        verify_claims = {
+            required
+            for required in self.deprecated_requires
+            if required in options
+        }
+        require_options = options.setdefault("require", [])
+        for opt in verify_claims:
+            opt_claim = opt.split("require_", 1)[1]
+            if options[opt]:
+                require_options.append(opt_claim)
+            warnings.warn(
+                "The {0} parameter is deprecated. Please add {1} to"
+                " the require list in options instead".format(opt, opt_claim),
+                DeprecationWarning,
+            )
+
         if isinstance(leeway, timedelta):
             leeway = leeway.total_seconds()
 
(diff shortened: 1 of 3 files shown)


jwt/api_jwt.py
```diff
@@ -100,6 +98,18 @@ def _validate_claims(self, payload, options, audience=None, issuer=None,
                           'Please use verify_exp in options instead.',
                           DeprecationWarning)
 
+        deprecated_requires = ['require_exp', 'require_iat', 'require_nbf']
+        verify_claims = set(deprecated_requires).intersection(options)
+        require_options = options.setdefault('require', [])
+        for opt in verify_claims:
+            opt_claim = opt[8:]
```
jpadilla: Took me a couple of seconds to understand what this was. Can we do something like `opt.split('require_', 1)[1]`?

jwt/api_jwt.py
```diff
@@ -100,6 +98,18 @@ def _validate_claims(self, payload, options, audience=None, issuer=None,
...
+            if options[opt]:
+                require_options.append(opt_claim)
+            warnings.warn('The ' + opt + 'parameter is deprecated. '
```
jpadilla: Let's do `warnings.warn('The {0} parameter is deprecated. Please add {1} to the require list in options instead.'.format(opt, opt_claim), DeprecationWarning)` instead

psafont:
Don't know why it's failing, hopefully I didn't mess the rebase.

jpadilla:
@psafont not to worry, that should be fixed once #297 is taken care of.

jpadilla:
@psafont mind rebasing from master? #297 is now solved.

psafont:
Just in case it was missed, I've rebased it again.

jwt/api_jwt.py
```diff
@@ -100,6 +98,18 @@ def _validate_claims(self, payload, options, audience=None, issuer=None,
                           'Please use verify_exp in options instead.',
                           DeprecationWarning)
 
+        deprecated_requires = ['require_exp', 'require_iat', 'require_nbf']
+        verify_claims = set(deprecated_requires).intersection(options)
```
mark-adams: It seems kind of silly to initialize a set here on every call. Why not initialize this set as a property on the `PyJWT` class so it happens once?

Also, are you sure doing a set intersection here is actually faster then simply doing something like:
```python
verify_claims = [requires for requires in deprecated_requires if requires in options]
```

In some cases (small sets of items), it may be faster to just do something like the above. I'm not saying it is in this case, but you should find out.

jwt/api_jwt.py
```diff
@@ -126,14 +136,9 @@ def _validate_claims(self, payload, options, audience=None, issuer=None,
...
-        if options.get('require_exp') and payload.get('exp') is None:
-            raise MissingRequiredClaimError('exp')
-
-        if options.get('require_iat') and payload.get('iat') is None:
-            raise MissingRequiredClaimError('iat')
-
-        if options.get('require_nbf') and payload.get('nbf') is None:
-            raise MissingRequiredClaimError('nbf')
+        for claim in options.setdefault('require', []):
+            if payload.get(claim) is None:
```
mark-adams: I think `options.get('require', [])` is a little more idiomatic than `options.setdefault('require', [])`.

psafont:
@mark-adams It's been quite a while since the last changes, is something else needed to accept these changes?

This functionality would still simplify my code :)

mathiasose:
Bump. I would also like to see this functionality released.

psafont:
Rebased patchset to master. Is there anything missing?

auvipy:
thanks for your patience!

PR #284, opened 2017-07-26
Include 'algorithms' kwarg in jwt.decode() (documentation)

vimalloc:
11f30c4050a11b6398d38f505578c9dabeba6c78 changes things so that a deprecation warning is raised when algorithms isn't specified in the jwt.decode(). This just updates the documentation to match. It also updates the first encode/decode example in the docs so you don't have to scroll a ways to the right in order to see that you need the `secret` and `algorithms` args.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -22,10 +22,11 @@ Example Usage
 
     >>> import jwt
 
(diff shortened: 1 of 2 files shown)


jpadilla:
@vimalloc nice, thanks!

PR #285, opened 2017-08-04
Require cryptography >= 1.4, replace deprecated functions

bluetech:
The signer/verifier interfaces are deprecated in cryptography==2.0. The
alternative sign/verify interfaces were added in cryptography==1.4.


mark-adams:
This looks good but it should be added to the CHANGELOG.

bluetech:
@mark-adams I added CHANGELOG entries.

mark-adams:
@bluetech Thanks for the contribution!

PR #289, opened 2017-09-05
Address over-eager fallback to stdin when decoding

milliburn:
On MacOS 10.12.6 using iTerm2 and Bash, the condition at https://github.com/jpadilla/pyjwt/blob/master/jwt/__main__.py#L57 appears to always evaluate to `True` regardless of whether the user had passed a JWT on the command line or not. Hence a command such as `pyjwt decode --no-verify json.web.token` (where `json.web.token` is a valid JWT) is not usable.

This change modifies the command-line interface such that the application will only attempt to read from stdin if the input had not been passed.

### DIFF
--- a/jwt/__main__.py
+++ b/jwt/__main__.py
@@ -54,10 +54,13 @@ def encode_payload(args):
 
 def decode_payload(args):
     try:
-        if sys.stdin.isatty():
-            token = sys.stdin.read()
-        else:
+        if args.token:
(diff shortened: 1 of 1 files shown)


jpadilla:
@milliburn thanks for getting this done. Confirmed this via #298 so just opened #304 to get this merged. Your original commit is still attributed to you.

PR #300, opened 2017-10-14
Currently validate iat only checks for iat type, not the date

vitoacme:
Here's the code:
```
    def _validate_iat(self, payload, now, leeway):
        try:
            int(payload['iat'])
        except ValueError:
            raise InvalidIssuedAtError('Issued At claim (iat) must be an integer.')
```

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -34,7 +34,7 @@ Exceptions
 
 .. class:: InvalidIssuedAtError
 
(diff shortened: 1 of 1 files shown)


lutzhorn:
Every integer can be treated as a timestamp and thus be converted to a `datetime`.

What validation would you like to happen except for the type?

Issue #297, opened 2017-10-04
Travis builds fail due to missing python2.6 and python3.3


sirosen:
@jpadilla, this is probably your call in terms of what lengths you want to go to in order to support these platforms, since it doesn't look like Travis py3.6 workers have these pythons anymore.

This happened in builds for #296, and it will impact other changes too.

### Drop 2.6?

If there's no testing on py2.6, pyjwt should drop the classifier for it: otherwise something like `"{}".format("foo")` will slip in, and on 2.6 it will break.
2.6 is _really old_ at this point. Is support for it important?

The last non-secuirty-fix release of 2.6 was almost a decade ago.

### Drop 3.3?

Python3 moves fast, and 3.3 is one of the older versions at this point.
It was released in 2012 and has been in security-fix-only maintenance since 2014 ([py3.3.7 release notes](https://www.python.org/downloads/release/python-337/)).

Unlike the issues with 2.6, if it works on 2.7 and 3.4, it _probably_ works on 3.3.
Still, IMO pyjwt should drop the classifier if it's not being tested.

### Work hard to test 2.6 and/or 3.3?

The other option seems to be to have Travis explicitly install these pythons for tests to run.
For the reasons noted above, I don't think this is worth the work involved.

If you wanted to do it, one of the better routes is probably to refactor the way tests are run to leverage Travis' python version selection, rather than trying to run everything through tox.

jpadilla:
I'm +1 on dropping 2.6 and 3.3 at this point.

jpadilla:
Closed via #301

PR #301, opened 2017-10-19
Attempt to drop support for python 2.6 and 3.3

vimalloc:
Per #297


jpadilla:
@vimalloc thanks for putting this together!

PR #302, opened 2017-10-26
Should not raise DecodeError('Invalid crypto padding') if verify = False

james4388:


jwt/api_jws.py
```diff
@@ -187,7 +187,8 @@ def _load(self, jwt):
         try:
             signature = base64url_decode(crypto_segment)
         except (TypeError, binascii.Error):
-            raise DecodeError('Invalid crypto padding')
+            if verify:
```
mark-adams: Instead of checking `verify` to silence the error, what do you think about not returning the signature at all if `verify=False`.

Maybe something like 

```
if not verify:
    return (payload, signing_inpiut, header, None)

try:
    signature = base64url_decode(crypto_segment)
except (TypeError, binascii.Error):
    raise DecodeError('Invalid crypto padding')
```

JonathanHuot: > Instead of checking `verify` to silence the error, what do you think about not returning the signature at all if `verify=False`.

I guess it is safest to `return (.., None)` *only* if it has an exception. Else, it should return the signature, that more logical and useful for the developer which uses the API.

Something like: 
```python
try:
    signature = base64url_decode(crypto_segment)
except (TypeError, binascii.Error):
    if verify:
        raise DecodeError('Invalid crypto padding')
    return (payload, signing_input, header, None)
```

Issue #144, opened 2015-04-19
Add support for JSON Web Key (JWK)


mark-adams:
http://datatracker.ietf.org/doc/draft-ietf-jose-json-web-key/

mark-adams:
This will cover the fix for #90 as well

bjmc:
I'm interested in this feature to use with Google's OpenIDConnect implementation, e.g. https://www.googleapis.com/oauth2/v3/certs

Is anyone working on this already, or should I go ahead and start a branch? Any tips for how you'd like to see this implemented?

Seraf:
Hello,

I also need this feature, is there any work done on it yet ?

mark-adams:
Not yet, the plan is to begin doing work on this against the 2.0-develop branch.

iiSeymour:
JWK is something I'd like to see in PyJWT :+1:

clintonb:
@mark-adams is there a timeframe for the release of this work and v2.0?

jpadilla:
Anyone interested in this should take a look at #202

bjmc:
In case anyone else needs this feature, it is supported in https://github.com/mpdavis/python-jose/

jpadilla:
Calling on @clintonb @iiSeymour @seraf and anyone else interested in JWK support to check out #202

lookuptable:
@jpadilla It seems the JWK support is on only when cryptography is available? Any chance that you'll support this feature with pycrypto so that it can be used in GAE environment?

mark-adams:
We don't actually support PyCrypto. We only kept it around for
compatibility since we didn't want to instantly break everyone using GAE
when we switched to cryptography.

I'd recommend that you look into the new GAE flexible environment which
will allow you to use any libraries you want including cryptography.

butla:
@mark-adams shouldn't this issue be closed? The code seems to be functional.

evertlammerts:
I've submitted #303 to support loading ECDSA keys from JWKs.

We use JWK sets to support key rotation and selecting a verify key based on the `kid` property in JWT headers. With the current API of PyJWT the user first needs to call `jwt.get_unverified_header`, grab the key from the keyset and call `jwt.decode`. This means base64 and json decoding the same data twice.

[...]

PR #303, opened 2017-10-31
from_jwk for ecdsa

evertlammerts:
Hi, here's an implementation of from_jwk for ecdsa keys. It DER encodes the given JWK and loads it with `cryptography.hazmat.primitives.serialization.load_der_*_key`. The patch works and has the basic tests in place. What do you think?

Also see my comment on [issue 144](https://github.com/jpadilla/pyjwt/issues/144#issuecomment-342273648)


evertlammerts:
I stupidly missed cryptography's ec.EllipticCurvePrivateNumbers and ec. EllipticCurvePublicNumbers. Closing this and creating a new PR using those instead.

PR #306, opened 2017-11-14
Allow list of valid audiences (#205 continued)

r-springer:
Extending @codervinod's work in #246 for #205 (since there's no progress there recently), addressing the PR feedback.

PR #308, opened 2017-11-20
Add support for checking against multiple issuers

jquense:
👋  hey there. This is a compliment to PR https://github.com/jpadilla/pyjwt/pull/306 it adds support for checking against multiple issuers. Our use-case is to enabling rolling migrations from one auth provider to another, where the audience but also the issuer has changed, without breaking verification for already issued tokens.

cc @yodaiken


taion:
Any update here? This seems like a pretty innocuous change, and it would simplify life for us.

PR #309, opened 2017-11-21
X5c jwk support

jquense:
Adds a small amount for support for certificates around jwk's along with small conveniences I'd like to see.

- the RSA algo now accepts a cert or a key in `to_jwk`
- adds the `x5c` and `x5t` keys to the jwk
- `from_jwk` will accept a dict and skip json deserializing if so (our jwks are already in memory)
- `from_jwk` uses the cert to return the public key if available.

This is mostly an attempt to upstream what we've been doing on top of pyJWT. thanks ya'll!

cc @taion @yodaiken


taion:
Any update here? I don't think there should be much in the way of downside to supporting the x5c bit, and it is a little bit more agnostic to the specific algorithm.

jpadilla:
LGTM. @mark-adams thoughts?

madrover:
I would love to have x5t support built in as well.
Could this PR be progressed?
Thanks!

PR #310, opened 2017-11-21
Add PyJWKSet

jquense:
Adds a new class for managing JWK sets. Let me know if this is moving in a direction you think makes sense. I can add comments and doc bits if this seems more or less like something you'd want to take

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,8 +7,8 @@ exclude =
 [wheel]
 universal = 1
 
-; [tool:pytest]
-; addopts = --cov-report term-missing --cov-config=.coveragerc --cov .
+[tool:pytest]
+addopts = --cov-report term-missing --cov-config=.coveragerc --cov .
 
 [aliases]
 test = pytest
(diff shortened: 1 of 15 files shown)


chirag04:
cc @jpadilla

setup.cfg
```diff
@@ -7,8 +7,8 @@ exclude =
 [wheel]
 universal = 1
 
-[tool:pytest]
-addopts = --cov-report term-missing --cov-config=.coveragerc --cov .
+; [tool:pytest]
+; addopts = --cov-report term-missing --cov-config=.coveragerc --cov .
```
wichert: Why are you commenting this out?

wichert:
Looking at the changes I don't see how one would use the set?

jquense:
i'm not sure i understand the question? Do the tests not demostrate usage, or you mean in more of a practical sense "I don't understand why someone would use this"?

jgehrcke:
I think this deserves a proper review. Superficially it looks nice.

PR #324, opened 2018-01-14
Fix hyperlink formatting in algorithms.rst

tornato7:
The hyperlink formatting was in Markdown while the document is in reStructuredText. Simply changing the formatting to match. Also making the link https. This is a security-oriented library, after all.


jpadilla:
thanks 🎉

PR #335, opened 2018-03-06
Expose InvalidSignatureError on jwt module

michael-k:
Regards jpadilla/pyjwt#316


jpadilla:
@michael-k great catch, thanks!

Issue #181, opened 2015-10-09
RSA256 not supported despite having cryptography (was: pycrypto) installed


wsw70:
Python 3.2.3 on a RPi (`Linux raspberrypi 3.18.11+ #781 PREEMPT Tue Apr 21 18:02:18 BST 2015 armv6l GNU/Linux`). 

PyCrypto is installed:

```
pycrypto                  - Cryptographic modules for Python.
  INSTALLED: 2.6.1 (latest)
```

`python3-dev` and `python3-crypto` are installed as well. Note: the same issue is present whether either of `pycrypto` or `python3-crypto` are installed, or both, or neither.

The following call crashes with an unsupported algorithm:

`myjwt = jwt.encode(claim, private_key, algorithm='RS256', headers={"alg": "RS256", "typ": "JWT"}).decode('utf-8')`

```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/local/lib/python3.2/dist-packages/jwt/api_jws.py", line 96, in encode
    alg_obj = self._algorithms[algorithm]
KeyError: 'RS256'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.2/threading.py", line 740, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.2/threading.py", line 693, in run
    self._target(*self._args, **self._kwargs)
  File "galarmclock.py", line 241, in getcalendar
    gauth.gettoken()
  File "/root/galarmclock/googleauth.py", line 32, in gettoken
    myjwt = jwt.encode(claim, private_key, algorithm='RS256', headers={"alg": "RS256", "typ": "JWT"}).decode('utf-8')
  File "/usr/local/lib/python3.2/dist-packages/jwt/api_jwt.py", line 56, in encode
    json_payload, key, algorithm, headers, json_encoder
  File "/usr/local/lib/python3.2/dist-packages/jwt/api_jws.py", line 101, in encode
    raise NotImplementedError('Algorithm not supported')
NotImplementedError: Algorithm not supported
```

mark-adams:
Thanks for reporting @wsw70.

One small thing first. You don't need to include the `headers` value in your call. Those headers will be automatically populated by PyJWT when it is creating the token. The `headers` parameter is for adding additional header parameters besides just those required by the spec.

[...]

PR #337, opened 2018-03-23
Update legacy instructions for using pycrypto

caseyduquettesc:
Related to #181 where users see `ValueError: Algorithm already has a handler.` after trying to register pycrypto algorithms. You must unregister the original handler first.


jpadilla:
@caseyduquettesc thanks for this! I now feel we should probably improve the experience around this, but this is a good first step.

PR #340, opened 2018-03-27
all exceptions inherit from PyJwtError

dradetsky:
I was trying to use pyjwt in some (connexion/flask) apps. I wanted to have all the jwt-related errors be converted into something (a `connexion.exceptions.ProblemException`  subclass) which my app could recognize as an auth failure & thus respond with 401. Maybe some of the exceptions should be treated as 5xx (I haven't actually tried to figure out exactly what they all mean yet), but nevermind.

It's not obvious from the API dox or even the code how to catch all the pyjwt exceptions. So I had what turn out to be the two base classes inherit from `PyJwtError` and made a comment explaining the purpose.

If this was my repo, I'd make the comment say "All exceptions _MUST_ inherit from `PyJwtError`" and reject any pr which adds an exception which doesn't, but it's not so that seems kinda impertinent.

### DIFF
--- a/jwt/exceptions.py
+++ b/jwt/exceptions.py
@@ -1,11 +1,11 @@
-class PyJwtError(Exception):
+class PyJWTError(Exception):
     """
     Base class for all exceptions
     """
     pass
 
 
-class InvalidTokenError(PyJwtError):
+class InvalidTokenError(PyJWTError):
     pass
 
 
@@ -37,7 +37,7 @@ class ImmatureSignatureError(InvalidTokenError):
     pass
 
 
-class InvalidKeyError(PyJwtError):
+class InvalidKeyError(PyJWTError):
     pass
 
 
--- a/jwt/exceptions.py
+++ b/jwt/exceptions.py
@@ -1,4 +1,11 @@
-class InvalidTokenError(Exception):
+class PyJwtError(Exception):
+    """
+    Base class for all exceptions
+    """
+    pass
+
+
+class InvalidTokenError(PyJwtError):
     pass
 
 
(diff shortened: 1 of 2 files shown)


jwt/exceptions.py
```diff
@@ -1,4 +1,11 @@
-class InvalidTokenError(Exception):
+class PyJwtError(Exception):
```
jpadilla: Small nitpick, can we rename this to `PyJWTError`? Also, we'll need to expose this in `pyjwt/__init__.py`.

mark-adams:
I'm not so sure I like this. All of our exceptions already inherit from a
single base exception. The name change just breaks the public API for no
real reason.

What about adding this exception and making InvalidTokenError inherit from
it?

mark-adams:
Whoops! The diff was not apparent from my email and I now see that it does
the EXACT thing I requested in my previous comment. 🤦‍♂️

jpadilla:
@mark-adams diff looks weird, but this is adding a new exception `PyJWTError` and updates `InvalidTokenError` to subclass from `PyJWTError` instead of `Exception`.

screen shot 2018-03-27 at 8 40 43 pm

mark-adams:
Yeah, the email experience was suboptimal here. My fault. LGTM.

jwt/exceptions.py
dradetsky: for you? of course

jpadilla:
@dradetsky thanks!

cactus:
Any chance of getting a release tagged (and uploaded to pypi) with this change in it, any time soon?

jpadilla:
@cactus v1.6.2 will be available shortly

PR #341, opened 2018-03-28
Add json_object_hook argument to PyJWT.decode

gvsheva:
Small change that allows decode object with custom types

### DIFF
--- a/jwt/api_jwt.py
+++ b/jwt/api_jwt.py
@@ -56,7 +56,7 @@ class PyJWT(PyJWS):
         )
 
     def decode(self, jwt, key='', verify=True, algorithms=None, options=None,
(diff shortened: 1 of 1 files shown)


mark-adams:
I think we tend to put options in two places: the `PyJWS` / `PyJWT` object constructors and the args to `encode()` / `decode()`. I think this is one of the cases where it probably makes sense to put it in the constructor to `PyJWS` / `PyJWT` since you would probably want to initialize a single instance with your custom config that is used multiple times.

gvsheva:
I've passed custom objects hook via `options` argument that there is in constructor and `decode()` method. So can you review again?

gvsheva:
@mark-adams, do I need doing something more?

Issue #312, opened 2017-11-27
Python type hints


mark-adams:
[PEP 484](https://www.python.org/dev/peps/pep-0484/) and [PEP 526](https://www.python.org/dev/peps/pep-0526/) introduced the concept of "type hints" which allow Python code to have inline type declarations to make it easier for editors and static analysis tools to work with Python code.

```python
def add(a: int, b: int) -> int:
    return a + b
```

To maintain Python 2.7 compatibility, we'd have to use the comment-based syntax:

```python
def add(a, b):
    # type: (int, int) -> int
    return a + b
```

Is this something that might be worthwhile for us to implement?

dmwyatt:
You can run mypy against the code and it will tell you if your types don't make sense.

jacopofar:
I'm happy to give it a try, and hopefully include mypy to travis config

jacopofar:
Opend PR #344

aklim007:
@jacopofar hello, in `decode` methods `jwt` and `jws` parameters may by bytes, not only str.

jdufresne:
Python type hints have since been added and are being improved upon. Perhaps it is okay to close this?

PR #344, opened 2018-04-06
RFC: Add type hints

jacopofar:
I decided to use the time my company, [Flixbus](https://www.flixbus.com/), dedicates to FOSS projects to contribute to this package.

This PR adds type hints as described in issue #312 and includes `mypy` in Travis configuration. This means that IDEs can automatically detect type errors and results from static check tools can improve.

There are a few doubts about which I'd like a feedback and for which I'm ready to work to change/fix:

- Mypy complains about `PyJWT.encode` not being a subtype of `PyJWS.encode`, and the same for `decode`. In the case of `decode` the problem is that one calls the token `jwt` and the other `jws`, so I just called both `token`. As for `encode` the problem is that `JWS` expects bytes and `JWT` a dictionary, so I changed `JWS.encode` to `JWS.encode_bytes`. In the documentation there are no examples about using that function, but it is a breaking change for who uses it.

- The change required to import the `typing` module in a `try...catch`, so it can be imported when checking the types but is ignored during normal usage of the library. It's a bit ugly but I couldn't find other ways to have it work on Python 2.7

- I didn't update the changelog or bump the version, if `JWS.encode` is considered part of the API I guess the major version number is affected


jwt/api_jws.py
```diff
@@ -69,8 +74,13 @@ def get_algorithms(self):
         """
         return list(self._valid_algs)
 
-    def encode(self, payload, key, algorithm='HS256', headers=None,
-               json_encoder=None):
+    def encode_bytes(self,
+                     payload,  # type: bytes
```
jpadilla: I'm wondering if we could leave this as `.encode()` and make the type for payload `Union[bytes, Dict]`?

jacopofar: It's possible, doing so mypy says:
 `jwt/api_jwt.py:57: error: Unsupported target for indexed assignment`
because the function is using the payload it as a dictionary. But before that there's a check for it and an error being thrown if it's not so I think it's fine to silence it.

The new commit does that and revert `encode_bytes` to `encode`

jwt/api_jwt.py
```diff
@@ -55,7 +66,12 @@ def encode(self, payload, key, algorithm='HS256', headers=None,
             json_payload, key, algorithm, headers, json_encoder
         )
 
-    def decode(self, jwt, key='', verify=True, algorithms=None, options=None,
```
aklim007: change jwt -> token, changes public interface and broken my code

jacopofar: Sorry, I didn't see it as a major change as the argument is positional, turns out I was wrong and there are users with the same problem. It has been reverted [here](https://github.com/jpadilla/pyjwt/issues/351)

PR #348, opened 2018-05-16
Updated usage doc to get header from token.

yks0000:


jpadilla:
Thanks! Going with #350 to address this.

PR #352, opened 2018-05-24
Fix #351 by reverting argument name changes for .decode()

### LINKED ISSUE #351: 1.6.1 --> 1.6.3 introduced breaking API changes: .decode(token=) vs .decode(jwt=)
PR #344 renamed some arguments on `api_jwt` ([for example](https://github.com/jpadilla/pyjwt/pull/344/files#diff-00ea5fc2260636f8d87435daf8c3b4c7R69)).  This is a breaking API change -- although the argument normalization is appreciated and an improvement, this should have triggered a major version bump (eg 1.6.1 --> 2.0.0) or there should have been a deprecation warning when using the old arg.

## Expected Result

[Semver](https://semver.org/) states:

> Given a version number MAJOR.MINOR.PATCH, increment the:
> MAJOR version when you make incompatible API changes,
> MINOR version when you add functionality in a backwards-compatible manner, and
> PATCH version when you make backwards-compatible bug fixes.

My dependency manager matches pyjwt on 1.6.*, so I expected a patch version bump to be not breaking.

Given the [wide usage](https://libraries.io/pypi/PyJWT/usage) of PyJWT, API should be stable across patch releases.

## Actual Result

My dependency manager pulled in 1.6.3, and suddenly my build broke.

mark-adams:
@jpadilla Can you give a quick review to this so we can get a new release out to fix this breaking API change?

### DIFF
--- a/jwt/api_jws.py
+++ b/jwt/api_jws.py
@@ -128,7 +128,7 @@ class PyJWS(object):
         return b'.'.join(segments)
 
     def decode(self,
-               token,  # type: str
+               jwt,  # type: str
                key='',   # type: str
                verify=True,  # type: bool
                algorithms=None,  # type: List[str]
@@ -146,7 +146,7 @@ class PyJWS(object):
                 DeprecationWarning
             )
 
-        payload, signing_input, header, signature = self._load(token)
+        payload, signing_input, header, signature = self._load(jwt)
 
         if not verify:
             warnings.warn('The verify parameter is deprecated. '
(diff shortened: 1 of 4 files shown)


jwt/api_jws.py
```diff
@@ -128,7 +128,7 @@ def encode(self,
         return b'.'.join(segments)
 
     def decode(self,
-               token,  # type: str
+               jwt,  # type: str
```
bogdanpetrea: @jpadilla, @mark-adams: I might be wrong, but this looks like it used to be `jws` before, as can be seen in here: https://github.com/jpadilla/pyjwt/commit/aed7305680b0ca6cf9846cd755b420f4dfccee22#diff-971ea3149c0d11912490c16fc9f982a9L120

PR #363, opened 2018-07-23
Fix #362 by add more invalid_strings with 'ecdsa-sha2'

### LINKED ISSUE #362: EC Signature allows to use ssh pub key format but HMAC doesn't check invalid secret key.
The ECDSA algorithm allows to use ssh public key format `ecdsa-sha2` but HMAC doesn't check invalid secret key. If users use ECDSA algorithm and they don't identify algorithm on decode, the attacker can tamper algorithm to HS. Then the server will use ECDSA public key with HMAC secret key.

bongtrop:


jpadilla:
@mark-adams any thoughts on this?

@bongtrop we should probably have some tests for this.

mark-adams:
@jpadilla Yea, the only real way to identify SSH public keys being used as HMAC secrets is by detecting strings like `ssh-rsa` or `ecdsa-sha2`. However, as long we we're here, we should also add `ssh-ed25519` and `ssh-dss` to the list since they can also be used in SSH public keys

Also, there should definitely be a test here.

bongtrop:
@mark-adams The reason that I put only the `ecdsa-sha2` to the restrict list is the PyJWT load only `ecdsa-sha2`. So are other public key formats necessary for current version?

bongtrop:
I'm no idea. Why does AppVeyor build failed? @jpadilla  @mark-adams  Please review!

jpadilla:
@bongtrop restarted AppVeyor build and it's passing now. We'd still need some tests for this.

randomstuff:
Any update on this? Can this be merged?

PR #368, opened 2018-08-14
Fixes example invocations in README.rst and index/docs.rst.

debayande:
The example invocations for decoding JWT tokens in the README appear to be incorrect. This PR, if applied, will fix the same.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -59,8 +59,8 @@ Usage::
 
 Decoding examples::
 
-    pyjwt --key=secret TOKEN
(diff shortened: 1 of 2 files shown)


jpadilla:
@debayande could you also update docs? https://github.com/jpadilla/pyjwt/blob/master/docs/index.rst

debayande:
@jpadilla Done!

jpadilla:
@debayande thanks!

PR #375, opened 2018-09-25
Import collection ABC's from correct module

kreutz-hs:
They were moved into collections.abc in 3.3 and will be deprecated
from collections in 3.8.


jaraco:
This fixes #381.

kreutz-hs:
So what does it mean? Somebody has apparently added tests that *require* that there will be a deprecation warning. ???
Those tests now fail this merge request.

Should I do something?

kreutz-hs:
OK, cool!  Thanks for all the work!

PR #378, opened 2018-10-11
from_jwk accepts dict

CDyWeb:
In many cases, the jwk is already a dict, so let's pass it on directly to from_jwk without flipping it back and forth from/to a json string

### DIFF
--- a/jwt/algorithms.py
+++ b/jwt/algorithms.py
@@ -162,7 +162,15 @@ class HMACAlgorithm(Algorithm):
 
     @staticmethod
     def from_jwk(jwk):
-        obj = json.loads(jwk)
+        try:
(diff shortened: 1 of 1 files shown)


jpadilla:
> could you plz add some tests to verify these changes?

would rather wait to include this in v2 https://github.com/jpadilla/pyjwt/pull/470/files#diff-dc58362670f457ab737771add140de08R292-R299

auvipy:
oh i see! i would focus on v2.0 now then

jpadilla:
Sneaked in here https://github.com/jpadilla/pyjwt/commit/dee2d31640940bfc77c3bc5839587a082eae58a4#diff-d76b80023c4cd94d656d2fc096a0e24e

PR #388, opened 2019-01-08
Fix case where payload audience is empty array

sarah-vanderlaan:
This PR fixes the case where an InvalidAudienceError is thrown when the application does not specify an audience but the token contains an audience that is an empty array.  Thanks 😄!

### DIFF
--- a/jwt/api_jwt.py
+++ b/jwt/api_jwt.py
@@ -175,7 +175,7 @@ class PyJWT(PyJWS):
             raise ExpiredSignatureError('Signature has expired')
 
     def _validate_aud(self, payload, audience):
-        if audience is None and 'aud' not in payload:
+        if audience is None and ('aud' not in payload or not payload['aud']):
             return
 
         if audience is not None and 'aud' not in payload:


jpadilla:
@sarah-vanderlaan thanks for putting this together.

I reproduced the issue by doing the following:

```python
import jwt

secret = 'shhhhh'
token = jwt.encode({'some': 'payload', 'aud': 'urn:foo'}, secret, algorithm='HS256')
print(token)
decoded = jwt.decode(token, secret, algorithms=['HS256'])
print(decoded)
```

Which results in:

```
b'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzb21lIjoicGF5bG9hZCIsImF1ZCI6InVybjpmb28ifQ.r_ww5-csXiaBvvCcJHi5fzNPXYfmZMBcZTpkwXKdBnI'
  File "python", line 6, in <module>
jwt.exceptions.InvalidAudienceError: Invalid audience
```

I also tried comparing the behavior to what [jsonwebtoken](https://github.com/auth0/node-jsonwebtoken) node package does

```javascript
var jwt = require('jsonwebtoken');
var secret = 'shhhhh'
var token = jwt.sign({ data: 'bar' }, secret, { audience: 'urn:foo' });
console.log(token);
var decoded = jwt.verify(token, secret);
console.log(decoded);
```

Which results in 

```
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJkYXRhIjoiYmFyIiwiaWF0IjoxNTQ3MDYwNTUxLCJhdWQiOiJ1cm46Zm9vIn0.07-JxWsZWD6GRpl3jKWkn3ap7dFNcTnLAjH__jk0XmY
{ data: 'bar', iat: 1547060551, aud: 'urn:foo' }
```

Would you mind adding a test to go together with this change?

jpadilla:
Actually, my comment isn't actually the same as to what you're describing, although very similar, my bad.

```
import jwt

secret = 'shhhhh'
token = jwt.encode({'some': 'payload', 'aud': []}, secret, algorithm='HS256')
print(token)
decoded = jwt.decode(token, secret, algorithms=['HS256'])
print(decoded)
```

```
b'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzb21lIjoicGF5bG9hZCIsImF1ZCI6W119.2-CXNANpC8dxnmgLUensniDKY_ryKa6DsnNyWgILfR8'
  File "python", line 6, in <module>
jwt.exceptions.InvalidAudienceError: Invalid audience
```

---

```
var jwt = require('jsonwebtoken');
var secret = 'shhhhh'
var token = jwt.sign({ data: 'bar' }, secret, { audience: [] });
console.log(token);
var decoded = jwt.verify(token, secret);
console.log(decoded);
```

```
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJkYXRhIjoiYmFyIiwiaWF0IjoxNTQ3MDcxMTE3LCJhdWQiOltdfQ.mfsM1BEmhMgcAarrkVz4w25R0FBYqafJoNdN5eFe-mU
{ data: 'bar', iat: 1547071117, aud: [] }
```

---

> 4.1.3.  "aud" (Audience) Claim
> 
>   The "aud" (audience) claim identifies the recipients that the JWT is
>   intended for.  Each principal intended to process the JWT MUST
>   identify itself with a value in the audience claim.  If the principal
>   processing the claim does not identify itself with a value in the
>   "aud" claim when this claim is present, then the JWT MUST be
>   rejected.  In the general case, the "aud" value is an array of case-
>   sensitive strings, each containing a StringOrURI value.  In the
>   special case when the JWT has one audience, the "aud" value MAY be a
>   single case-sensitive string containing a StringOrURI value.  The
>   interpretation of audience values is generally application specific.
>   Use of this claim is OPTIONAL.

After reading this, I think this change is ok. Anyone else care to jump in?

Issue #392, opened 2019-01-24
Confused with jwt.decode return type


Hanaasagi:
I am using mypy to check my code. And I find that the annotation  of `jwt.decode`'s  return value is `Mapping[str, Any]`. It's immutable, so if i change it, mypy will complain `Unsupported target for indexed assignment`. I think the type should be `Dict[str, Any]` or `MutableMapping[str, Any]`. Are there someone give a official solution?

More detail, please see https://github.com/python/typeshed/issues/2764

jacopofar:
Seems a good idea, I tried to[ implement it](https://github.com/jacopofar/pyjwt/commit/4799c4e2cb462c4ef2db89ab81779afc344b69fd) and it passes the tests and mypy.

It took me some time to realize why it was `Mapping`, and apparently mypy inferred it from the check [at line 100](https://github.com/jacopofar/pyjwt/blob/decode-return-dict/jwt/api_jwt.py#L100).

Hanaasagi:
After `json.loads`, the key of `dict` must be a `str` type, maybe `Dict[str, Any]` is better. I've made some changes in https://github.com/python/typeshed/pull/2765.

jacopofar:
Right, fixed. Is it still needed?

jpadilla:
@jacopofar yes please! 👍

PR #393, opened 2019-01-27
Decode return type is dict[str, Any]

jacopofar:
As discussed [here](https://github.com/jpadilla/pyjwt/issues/392) the return type for `decode` should be `dict[str, Any]` instead of the (inferred) `Mapping[any, any]`, this adds it.

Mapping is an ABC and I think the check with `isinstance` maybe was more appropriate, but still we know it is going to be a dictionary in any case so should be fine.


Hanaasagi:
resolves #392

jpadilla:
@jacopofar @Hanaasagi thank you

PR #411, opened 2019-04-25
Make jwt consistent with different environment(language)

bitwolaiye:
The dict is inconsistent, some are in hash order, and some are in the order of definition. The inconsistent order of the dict will cause inconsistencies in all aspects of jwt, which is inconvenient to pass in different systems and implementations.

1. add order_dict to confirm dict with agreed order
2. fix problem: the same argument with different result between python2 and python3 (because python2 dict will reorder with hash order)
3. order payload

### DIFF
--- a/jwt/api_jws.py
+++ b/jwt/api_jws.py
@@ -15,7 +15,7 @@ from .exceptions import (
     DecodeError, InvalidAlgorithmError, InvalidSignatureError,
     InvalidTokenError
 )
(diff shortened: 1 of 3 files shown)


auvipy (changes requested):
would you mind resolve the conflicts?

jpadilla:
> The inconsistent order of the dict will cause inconsistencies in all aspects of jwt, which is inconvenient to pass in different systems and implementations.

Are there any concrete example/scenarios where this is true?

Issue #412, opened 2019-04-30
Error decoding on command-line: Expected a string value


jaraco:
Running a simple `decode` operation on the command line fails on Python 3:

```
$ pyjwt decode                                                                                                                                    
eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VyX2lkIjoxMjM0NTYsIm5hbWUiOiJcImphcmFjbyBkZXZcIiIsImFkbWluIjp0cnVlfQ.sx1EqsdNkZ2IAkDGmqs6PCjFPG0XNvtAYvlMXmyIG4g
There was an unforseen error:  Expected a string value
usage: 
    Encodes or decodes JSON Web Tokens based on input.

...
```

jaraco:
Aha. The issue is that either `--key` or `--no-verify` must be passed, but if it's not, it triggers the reported error. Allowing the traceback to propagate, it's more clear what's happening:

```
Traceback (most recent call last):
  File ".tox/python/bin/pyjwt", line 10, in <module>
    sys.exit(main())
  File "/Users/jaraco/code/public/pyjwt/.tox/python/lib/python3.7/site-packages/jwt/__main__.py", line 162, in main
    output = arguments.func(arguments)
  File "/Users/jaraco/code/public/pyjwt/.tox/python/lib/python3.7/site-packages/jwt/__main__.py", line 65, in decode_payload
    data = decode(token, key=args.key, verify=args.verify)
  File "/Users/jaraco/code/public/pyjwt/.tox/python/lib/python3.7/site-packages/jwt/api_jwt.py", line 93, in decode
    jwt, key=key, algorithms=algorithms, options=options, **kwargs
  File "/Users/jaraco/code/public/pyjwt/.tox/python/lib/python3.7/site-packages/jwt/api_jws.py", line 156, in decode
    key, algorithms)
  File "/Users/jaraco/code/public/pyjwt/.tox/python/lib/python3.7/site-packages/jwt/api_jws.py", line 220, in _verify_signature
    key = alg_obj.prepare_key(key)
  File "/Users/jaraco/code/public/pyjwt/.tox/python/lib/python3.7/site-packages/jwt/algorithms.py", line 140, in prepare_key
    key = force_bytes(key)
  File "/Users/jaraco/code/public/pyjwt/.tox/python/lib/python3.7/site-packages/jwt/utils.py", line 30, in force_bytes
    raise TypeError('Expected a string value')
[...]

PR #413, opened 2019-04-30
Get token from command-line or from stdin, irrespective of an interactive terminal.

### LINKED ISSUE #410: Why not read when not a TTY?
I was looking to experiment with PyJWT's command-line interface, and I thought a thing to try would be to encode a token and then decode it immediately, so I did this:

```
~ $ pyjwt --key='foo' encode user_id=123456 name="jaraco dev" admin=true | pyjwt decode                                                                                        
There was an unforseen error:  Cannot read from stdin: terminal not a TTY
usage: 
...
```

But as you can see, that generated an error. Is there a reason to reject the stdin if it's not a TTY?

jaraco:
Fixes #410 (presumably). Untested due to #412.

### DIFF
--- a/jwt/__main__.py
+++ b/jwt/__main__.py
@@ -54,14 +54,7 @@ def encode_payload(args):
 
 def decode_payload(args):
     try:
-        if args.token:
-            token = args.token
-        else:
-            if sys.stdin.isatty():
-                token = sys.stdin.readline().strip()
-            else:
-                raise IOError('Cannot read from stdin: terminal not a TTY')
-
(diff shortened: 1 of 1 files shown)


jaraco:
Figuring out the correct command syntax, I'm able to verify the intended behavior:

```
$ pyjwt --key='foo' encode user_id=123456 name="jaraco-dev" admin=true | .tox/python/bin/pyjwt decode --no-verify                                 
{"user_id": 123456, "name": "\"jaraco-dev\"", "admin": true}
```

PR #417, opened 2019-04-30
Invoke pytest directly.

### LINKED ISSUE #416: Package uses easy_install for tests
The setuptools project has deprecated use of `setup_requires` and `tests_require`, because they rely on `easy_install` and `eggs`, both of which are deprecated. Instead of using `setup.py test` this project should invoke pytest directly.

jaraco:
Avoids use of easy_install and interaction between flake8 and tests. Fixes #416.


jpadilla:
@jaraco thanks!

Issue #429, opened 2019-07-03
TypeError: sign() takes 3 positional arguments but 4 were given


PiotrNestor:
I'm getting:
TypeError: sign() takes 3 positional arguments but 4 were given

## Expected Result

Token is returned

## Actual Result

TypeError: sign() takes 3 positional arguments but 4 were given

```
   File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/jwt/api_jwt.py", line 65, in encode
    json_payload, key, algorithm, headers, json_encoder
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/jwt/api_jws.py", line 114, in encode
    signature = alg_obj.sign(signing_input, key)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/jwt/algorithms.py", line 313, in sign
    return key.sign(msg, padding.PKCS1v15(), self.hash_alg())
TypeError: sign() takes 3 positional arguments but 4 were given
```

## Reproduction Steps

```python
import jwt

private_key = b'''-----BEGIN EC PRIVATE KEY-----
...
-----END EC PRIVATE KEY-----'''
claim2 = { "Keyid": "abcdefg"}
token = jwt.encode(claim2, private_key, algorithm='RS256')
```

## System Information

    $ python -m jwt.help

```
{
  "cryptography": {
    "version": "2.7"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.7.3"
  },
  "platform": {
    "release": "18.6.0",
    "system": "Darwin"
  },
  "pyjwt": {
    "version": "1.7.1"
  }
}
```

This command is only available on PyJWT v1.6.3 and greater. Otherwise,
please provide some basic information about your system.

PiotrNestor:
This is on MacOS:
uname -a
Darwin EU-EXT-CKY 18.6.0 Darwin Kernel Version 18.6.0: Thu Apr 25 23:16:27 PDT 2019; root:xnu-4903.261.4~2/RELEASE_X86_64 x86_64

drbitboy:
Hi Piotr et al.,

**Summary**

The ```Reproduction Steps``` example has a mismatch between the Elliptic Curve ```private_key``` argument and the ```algorithm='RS256'``` keyword argument, which mismatch generates an exception; that exception, though less than helpful, is correct behavior.

**Suggestion**

[...]

PR #431, opened 2019-07-16
Improve exception clarity on key mismatch e.g. see issue #429 on Github

drbitboy:


jpadilla:
Thanks! There are some tests failing, but we'd also need some tests that cover these new improvements.

PR #438, opened 2019-09-28
Correct type for json_encoder argument

jdufresne:
Per recent upstream fix to typeshed, json.dumps() cls argument should be
optional type JSONEncoder.

https://github.com/python/typeshed/commit/8e0d288ea49a34f9bd21b1598ec487414a339a1f

Fixes mypy error:

    jwt/api_jws.py:102: error: Argument "cls" to "dumps" has incompatible type "Optional[Callable[..., Any]]"; expected "Optional[Type[JSONEncoder]]"
    jwt/api_jwt.py:61: error: Argument "cls" to "dumps" has incompatible type "Optional[Callable[..., Any]]"; expected "Optional[Type[JSONEncoder]]"

### DIFF
--- a/jwt/api_jws.py
+++ b/jwt/api_jws.py
@@ -3,7 +3,7 @@ import json
 import warnings
 try:
     # import required by mypy to perform type checking, not used for normal execution
-    from typing import Callable, Dict, List, Optional, Union # NOQA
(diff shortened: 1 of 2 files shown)


jpadilla:
Thanks!

PR #442, opened 2019-09-28
Remove setup.py test command in favor of pytest and tox

### LINKED ISSUE #415: flake8 fails after running tests
Running tests with `tox -e python` causes `setup.py pytest` to be invoked, which causes eggs to be installed in `./.eggs`. Thereafter, running `tox -e flake8` fails because the linter is run on everything, including files in `.eggs`. Probably the linter should be configured not to run on files in .eggs.

jdufresne:
Using pytest and tox directly is simpler and more conventional then
going through setup.py. Using setup.py installs packages as eggs where
as tox uses the more typical pip.

Overall simplifies setup.py.

Fixes #415

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,6 +9,3 @@ universal = 1
 
 [tool:pytest]
 addopts = --cov-report term-missing --cov-config=.coveragerc --cov .
-
-[aliases]
-test = pytest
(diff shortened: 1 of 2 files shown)


jpadilla:
Perfect, thanks! 🎆

PR #446, opened 2019-10-17
Don't accept decode(..., algorithm=...)

JulienPalard:
The `api_jws.decode()` function was accepting **kwargs but not using it. From my point of view, it's a lie: you can call `decode` with a typo in a parameter without any warning.

The bad thing is one could do:

    jwt.decode(token, pub_key, algorithm="ES256")

without noticing the `algorithm` is ignored, thus opening a vulnerabiliy.

By not accepting the **kwargs, we're enforing all parameters are "eaten" by the previous class, leading to a nice explicit:

    TypeError: decode() got an unexpected keyword argument 'algorithm'

error being raised.

Tried to run tox but I was having:

    jwt/help.py:12: error: Incompatible types in assignment (expression has type "None", variable has type Module)

before my patch, so it looks like my patch introduced no new errors.

I had to move `**kwargs` out of `_validate_claims`, it should not really change its signature, except that know it does no longer accept verify_expiration directly (but as it's a "private" method, I think it's OK).

I also had to move kwargs poping before calling `super().decode(`, so it won't get unexpected parameters.

### DIFF
--- a/jwt/api_jws.py
+++ b/jwt/api_jws.py
@@ -132,7 +132,7 @@ class PyJWS(object):
                verify=True,  # type: bool
                algorithms=None,  # type: List[str]
                options=None,  # type: Dict
-               **kwargs):
+               ):
 
         merged_options = merge_dict(self.options, options)
         verify_signature = merged_options['verify_signature']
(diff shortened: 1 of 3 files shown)


JulienPalard:
> Overall coverage decreased (-8.2%) to 68.188%

Not sure what to do with this, I only touched `jwt/api_jw*.py` and according to local tests, their coverage has not changed.

For the travis failures, I'm not sure to understand them.

akx:
This was revisited by #657 but it's a semver-major change (see #701).

PR #448, opened 2019-10-18
Catch BadSignatureError raised by ecdsa 0.13.3 on verification errors

### LINKED ISSUE #447: python-ecdsa 0.13.3 raises BadSignatureError exception
ECAlgorithm.verify expects an AssertionError when the signature is too long:
https://github.com/jpadilla/pyjwt/blob/3a20892442b34c754b26550e05f7f856fb008c94/jwt/contrib/algorithms/py_ecdsa.py#L59

But ecdsa>=0.13.3 raise an MalformedSignature exception internally, which is caught and reraised as BadSignatureError.

https://github.com/warner/python-ecdsa/blob/5fa2fd85aa4706c809e3ab705f3a22bf604eec24/src/ecdsa/keys.py#L521

## Expected Result

Catch BadSignature error, return false.

StefanBruens:
The new ecdsa no longer uses AssertionError when the signature is too long.
This happens in the test suite, where "123" is appended to the signature.

Fixes #447

### DIFF
--- a/jwt/contrib/algorithms/py_ecdsa.py
+++ b/jwt/contrib/algorithms/py_ecdsa.py
@@ -56,5 +56,7 @@ class ECAlgorithm(Algorithm):
         try:
             return key.verify(sig, msg, hashfunc=self.hash_alg,
                               sigdecode=ecdsa.util.sigdecode_string)
-        except AssertionError:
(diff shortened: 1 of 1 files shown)


jpadilla:
@StefanBruens great catch, thanks!

PR #456, opened 2019-11-03
Improve requiring claims

dave-shawley:
Thanks for implementing and maintaining this library!  This PR changes the processing of required claims so that arbitrary claims may be required by name.  I needed to require that the issuer claim was present without requiring a specific value.  There wasn't a easy way to do this, so I figured that I would add one.

I converted the required claim processing to look at any option starting with `require_` and require that a claim named by the suffix is present in the payload.  Hopefully that explanation makes some sense.  If not, the code is pretty simple.

I also took a stab at fixing the testing against Pycryptodome.  I can remove the test related commits if necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,11 @@ language: python
 matrix:
   include:
     - python: 3.5
-      env: TOXENV=py35-crypto,py35-nocrypto,py35-contrib_crypto
+      env: TOXENV=py35-crypto,py35-nocrypto,py35-contrib_crypto,py35-contrib_cryptodome
     - python: 3.6
-      env: TOXENV=py36-crypto,py36-nocrypto,py36-contrib_crypto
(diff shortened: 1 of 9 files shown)


dave-shawley:
@jpadilla not sure what to do about the typing failure.  I'm assuming that it does not like the `try`-`except AttributeError` to detect whether we should be using `Crypto.PublicKey.RSA._RSAobj` or `Crypto.PublicKey.RSA.RsaKey` but I'm not completely sure what to do about it.

dave-shawley:
@jpadilla 
I didn't notice #280 which implements similar functionality in a different manner.  Is that approach preferred over the one that I have taken in this PR?

I am also bewildered by the travis build failure ...

dave-shawley:
Looks like #280 was merged 👍

Issue #453, opened 2019-10-22
Python 2 support ?


Djailla:
Hello,

Python 2 support is almost over, do you think you should remove this support in the next major release ? It would allow extensive usage of mypy and f-strings and facilitate compatibility checks.

If agreed, I could provide a PR.

Regards

KingDarBoja:
I do think the next major release should remove it as stated on [this anaconda post](https://www.anaconda.com/end-of-life-eol-for-python-2-7-is-coming-are-you-ready/).

Most packages already participating on it as seen [here](https://python3statement.org/).

Cheers!

michael-k:
@Djailla can you close this issue?

PR #457, opened 2019-11-04
Remove Python 2.7 compatibility

Djailla:
Python 2.7 will be deprecated beginning 2020, remove support

Related to #453


Djailla:
Hello @jpadilla what should I do next with your approval ?

Issue #468, opened 2019-12-16
pyJWT does not support lists 


anguelov:
Summary.
A empty list or a list of JSON objects is valid JSON. currently trying to encode a list into JWT raises a exception. 
## Expected Result

`JWT.encode([], "secret", algorithm="HS256")` should return a valid JWT:
`eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.W10.ny0W7Pkele2S9JjQz1pcPuE-7_Jrt2VBqPEO1UnPyQg`

## Actual Result

```python
self = <jwt.api_jwt.PyJWT object at 0x1108f14d0>, payload = []
key = 'secret'
algorithm = 'HS256', headers = None, json_encoder = None

    def encode(self,
               payload,  # type: Union[Dict, bytes]
               key,  # type: str
               algorithm='HS256',  # type: str
               headers=None,  # type: Optional[Dict]
               json_encoder=None  # type: Optional[Callable]
               ):
        # Check that we get a mapping
        if not isinstance(payload, Mapping):
>           raise TypeError('Expecting a mapping object, as JWT only supports '
                            'JSON objects as payloads.')
E           TypeError: Expecting a mapping object, as JWT only supports JSON objects as payloads.
```

## Reproduction Steps

```python
import jwt
jwt.encode([], "secret")  # fail
jwt.encode([{}], "secret")  # fail
```

## System Information

    $ python -m jwt.help

```
{
  "cryptography": {
    "version": ""
  },
  "implementation": {
    "name": "CPython",
    "version": "3.7.5"
  },
  "platform": {
    "release": "19.0.0",
    "system": "Darwin"
  },
  "pyjwt": {
    "version": "1.7.1"
  }
}

```

jpadilla:
Not sure this is something we'll want to do though. According to the spec:

> JWTs represent a set of claims as a JSON object that is encoded in a
   JWS and/or JWE structure.  This JSON object is the JWT Claims Set.
   As per Section 4 of RFC 7159 [RFC7159], the JSON object consists of
   zero or more name/value pairs (or members), where the names are
   strings and the values are arbitrary JSON values.

PR #469, opened 2019-12-24
#468 | Added support of lists

JHJohny:
As in #468 was mentioned, list is valid JSON object (empty or non empty)
So I added support of lists.
- Edited encode and decode function
- Removed lists from test as bad type

Result - 
```
>>> import jwt
>>> 
>>> encoded = jwt.encode([], 'secret', algorithm='HS256')
>>> jwt.decode(encoded, 'secret', algorithms=['HS256'])
[]

>>> encoded = jwt.encode(["some", "ranom", "list", "values"], 'secret', algorithm='HS256')
>>> jwt.decode(encoded, 'secret', algorithms=['HS256'])
['some', 'ranom', 'list', 'values']
```

### DIFF
--- a/jwt/api_jwt.py
+++ b/jwt/api_jwt.py
@@ -112,7 +112,7 @@ class PyJWT(PyJWS):
             payload = json.loads(decoded.decode("utf-8"))
         except ValueError as e:
             raise DecodeError("Invalid payload string: %s" % e)
-        if not isinstance(payload, dict):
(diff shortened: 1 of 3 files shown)


jpadilla:
Hey @JHJohny, thanks for the contribution, although I'm not sure this is something we'll want to do. See https://github.com/jpadilla/pyjwt/issues/468#issuecomment-569291799 for more details.

PR #470, opened 2019-12-27
Version 2.0.0

jpadilla:
# Changes

## Introduce better experience for JWKs

Introduce `PyJWK`, `PyJWKSet`, and `PyJWKClient`.

TODO: Allow passing a Request Session object to `PyJWKClient`.

```python
import jwt
from jwt import PyJWKClient

token = "eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiIsImtpZCI6Ik5FRTFRVVJCT1RNNE16STVSa0ZETlRZeE9UVTFNRGcyT0Rnd1EwVXpNVGsxUWpZeVJrUkZRdyJ9.eyJpc3MiOiJodHRwczovL2Rldi04N2V2eDlydS5hdXRoMC5jb20vIiwic3ViIjoiYVc0Q2NhNzl4UmVMV1V6MGFFMkg2a0QwTzNjWEJWdENAY2xpZW50cyIsImF1ZCI6Imh0dHBzOi8vZXhwZW5zZXMtYXBpIiwiaWF0IjoxNTcyMDA2OTU0LCJleHAiOjE1NzIwMDY5NjQsImF6cCI6ImFXNENjYTc5eFJlTFdVejBhRTJINmtEME8zY1hCVnRDIiwiZ3R5IjoiY2xpZW50LWNyZWRlbnRpYWxzIn0.PUxE7xn52aTCohGiWoSdMBZGiYAHwE5FYie0Y1qUT68IHSTXwXVd6hn02HTah6epvHHVKA2FqcFZ4GGv5VTHEvYpeggiiZMgbxFrmTEY0csL6VNkX1eaJGcuehwQCRBKRLL3zKmA5IKGy5GeUnIbpPHLHDxr-GXvgFzsdsyWlVQvPX2xjeaQ217r2PtxDeqjlf66UYl6oY6AqNS8DH3iryCvIfCcybRZkc_hdy-6ZMoKT6Piijvk_aXdm7-QQqKJFHLuEqrVSOuBqqiNfVrG27QzAPuPOxvfXTVLXL2jek5meH6n-VWgrBdoMFH93QEszEDowDAEhQPHVs0xj7SIzA"
kid = "NEE1QURBOTM4MzI5RkFDNTYxOTU1MDg2ODgwQ0UzMTk1QjYyRkRFQw"
url = "https://dev-87evx9ru.auth0.com/.well-known/jwks.json"

jwks_client = PyJWKClient(url)
[...]

PR #475, opened 2020-02-21
Only parse jwk data as json when passing in a string

belak:
This allows a user to request the `/.well-known/jwks.json` file from a server, parse it, look up the appropriate key, and load it without having to dump the key to json again.

I realize this will be superseded by version 2.0 as that introduces better support for JWKs in general, but this fixes the given use case for us until that releases.

### DIFF
--- a/jwt/algorithms.py
+++ b/jwt/algorithms.py
@@ -121,14 +121,14 @@ class Algorithm(object):
     @staticmethod
     def to_jwk(key_obj):
         """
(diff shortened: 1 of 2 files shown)


belak:
It looks like the only failing build is related to python 3.8 support with PyCrypto (due to time.clock() being removed). It's the same failure as master.

belak:
This is most likely superceded by #470 which greatly improves on the JWK support.

PR #479, opened 2020-03-28
Custom JSON Decoder

eddloschi:
Added support for a custom JSON decoder class when decoding a jwt


auvipy (changes requested):
Could u please re base?

PR #482, opened 2020-04-24
Document top-level .encode and .decode to close #459

### LINKED ISSUE #459: Better documentation: complete encode and decode docs
https://pyjwt.readthedocs.io/en/latest/index.html has examples 😄
https://pyjwt.readthedocs.io/en/latest/api.html has only exceptions 😕

I wish there was complete documentation for `jwt.encode` and `jwt.decode` high level methods / public API.

Specifically, there are keyword arguments which are not trivial and `options=` which have defaults, but it takes digging into the source code to understand how the earlier work and what the latter are.

dimaqq:
Document `jwt.encode` and `jwt.decode` and close #459 (maybe, I mean, the classes need to be documented too, but maybe that's not as important).

Also clean up rst heading markup because shpinx complained about that.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -3,7 +3,67 @@ API Reference
 
 .. module:: jwt
 
-TODO: Document PyJWS / PyJWT classes
+.. function:: encode(payload, key, algorithm="HS256", headers=None, json_encoder=None)
+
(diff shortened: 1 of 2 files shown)


jpadilla:
this is great, thanks!

PR #484, opened 2020-04-24
Improve documentation for audience usage

CorreyL:
Hello `pyjwt` maintainers,

I've recently been using the `pyjwt` codebase to decode JWT tokens originating from [Keycloak](https://www.keycloak.org/) servers, where the JWT `aud` claim is an array by default, and I'm dealing with a `python` `API` server that supports multiple `audiences`.

The edits I've made in this PR hope to improve the documentation for users who are not as familiar with the specific details surrounding the `audience` claim, like I was until I started digging into using `pyjwt` more.

#### ddc1dc1

Makes an explicit code block example for the `aud` claim being an `array`. It was mentioned in the text, but I think making a code block demonstrating it helps make it more clear.

#### 46d1940

Adds an example of the `audience` param for `jwt.decode` supporting an iterable argument, as implemented in https://github.com/jpadilla/pyjwt/pull/306.

Note that I'm not stuck to the placement of this example, but I was unsuccessful at locating documentation specific to the `jwt.decode` implementation

#### f01df86

Is a simple `.rst` formatting change where the linter I was using was throwing a warning:

image

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -211,6 +211,19 @@ a single case-sensitive string containing a StringOrURI value.
     token = jwt.encode(payload, 'secret')
     decoded = jwt.decode(token, 'secret', audience='urn:foo', algorithms=['HS256'])
 
+If multiple audiences are accepted, the ``audience`` parameter for
+``jwt.decode`` can also be an iterable
+
+.. code-block:: python
+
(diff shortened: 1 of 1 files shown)


jpadilla:
this is great, thanks!

PR #490, opened 2020-06-07
Fix `tox -e lint` warnings and errors

jdufresne:
Now successfully passes when running `tox -e lint`.


jdufresne:
Rebased :slightly_smiling_face:

PR #495, opened 2020-06-11
Move setup information to declarative setup.cfg

jdufresne:
Use a declarative syntax to avoid mixing code and configuration.
Simplifies handling of long description and version by reducing some
boilerplate.

For details on this setuptools feature, see:
https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files


setup.cfg
```diff
@@ -0,0 +1,49 @@
...
+console_scripts =
+    pyjwt = jwt.__main__:main
+
+[options.extras_require]
+crypto =
+    cryptography >= 1.4
+tests =
+    pytest>=4.0.1,<5.0.0
+    pytest-cov>=2.6.0,<3.0.0
+
```
jpadilla, line 45: @jdufresne is it possible to do something like this using this setup?

```
EXTRAS_REQUIRE["dev"] = (
    EXTRAS_REQUIRE["tests"]
    + EXTRAS_REQUIRE["cryptography"]
    + EXTRAS_REQUIRE["jwks-client"]
    + ["mypy", "pre-commit"]
)
```

jdufresne: As far as I know, values can't be added in `setup.cfg` like in Python code, but we could add a `dev` extras and just copy the entries manually.

PR #513, opened 2020-08-22
Return tokens as string not bytes

jpadilla:
Extracted from #470

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -41,7 +41,8 @@ Usage
 
     >>> import jwt
     >>> encoded = jwt.encode({'some': 'payload'}, 'secret', algorithm='HS256')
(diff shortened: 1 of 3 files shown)


auvipy (approved):

mato2000:
Hello, I would like to know why tokens was changed to be string instead of bytes. It seems to me that it is more suitable to be bytes. Thanks

personalcomputer:
Hi @jpadilla, thanks for all your work on PyJWT. Just as a quick bit a feedback, I want to let you know that this return type 
change from bytes to str is a breaking change. Updating to 2.0.0 caused all our code interacting with `encode()` to fail, as our  code has been expecting bytes in return. It was hard to identify what happened as the release notes had listed this item embedded alongside all the other internal changes. It would be helpful that when publishing a new release, to prominently mention *breaking* API changes in the release notes, so that we can easily see them and update our code to support the changed interface when updating the library. Not required though, just thought I would let you know.

Issue #487, opened 2020-05-22
validate nbf uses local timezone for its current time. I believe it should use UTC


matt-coppola:
Summary.
_validate_nbf uses the local timezone for now when verifying whether the token is valid but should be using UTC time.

## Expected Result

The time validation should be using UTC.

## Actual Result

The validation uses the local timezone when determining the current time. This occurs in jwt/api_jwt.py ~line 154 when the call to timegm() converts to calendar time. Because I am in EST the time now is earlier than the nbf time set and thus validation fails. The NBF time was set to UTC time. 

I think this is a larger consistency issue at play. Any token that was created in a different timezone than that token is being decoded in will have varying results on whether the token is valid or not. 

## Reproduction Steps

1. Change your system timezone to be something earlier than GMT (Mine is EST)
2. Create a JWT with it's nbf set to the current UTC time (datetime.datetime.utcnow()).
3. Attempt to decode the token with validation for nbf. The validation will fail because the time now is earlier than the NBF time.

## System Information

    $ python -m jwt.help

```
{
  "cryptography": {
    "version": "2.5"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.8.2"
  },
  "platform": {
    "release": "5.4.0-25-generic",
    "system": "Linux"
  },
  "pyjwt": {
    "version": "1.7.1"
  }
}
```

auvipy:
any failing testcase?

matt-coppola:
Not a failing unit test. I believe the reason you won't see any failures in UT is because the test is doing both the encoding and decoding. Given that it will be doing this UT on the same machine (with the same system timezone) you won't observe this error. The issue is that decode assumes the current system timezone is the same as the timezone the token was encoded in.

matt-coppola:
Some additional details.
From the spec. https://tools.ietf.org/html/rfc7519 
An NBF must be a NumericDate https://tools.ietf.org/html/rfc7519#section-4.1.5 and a NumericDate must be in UTC https://tools.ietf.org/html/rfc7519#section-2

If you manually were to create a token with the NBF set to a utc time, and then tried to decode the token in a timezone such as EST the verification would fail because the time we are comparing against has been converted to EST by the call to timegm

m10d:
I believe i encountered the same underlying issue upgrading an old project.

note there is likely a bug in decode today; handling of timezones is broken if the jwt is not encoded in a UTC localtime.

```    
[...]

PR #516, opened 2020-09-05
Add decode timezone support

guyzyl:
Added `timezone` support to `options` dictionary.
The change is needed since the expiry of the tokens are validated against the machines local timezone, this new option enables explicitly choosing and setting the timezone used for decoding and validations.
I made to sure to document the new option + wrote a test for it.
This solves issue #487 .

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -51,6 +51,7 @@ API Reference
             .. deprecated:: 1.2.0
                 Use ``verify_exp`` instead
 
+        * ``timezone`` change the timezone used for decoding and validation, expects tzinfo object
 
     :param iterable audience: optional, the value for ``verify_aud`` check
     :param str issuer: optional, the value for ``verify_iss`` check
--- a/jwt/api_jwt.py
+++ b/jwt/api_jwt.py
@@ -36,7 +36,7 @@ class PyJWT(PyJWS):
             "verify_iat": True,
             "verify_aud": True,
             "verify_iss": True,
-            "timezone": timezone.utc,
+            "timezone": timezone.utc,   # type: tzinfo
             "require": [],
         }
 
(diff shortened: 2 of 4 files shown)


auvipy (approved):

jpadilla:
@auvipy we need to think through this one. what does the spec say? what are other libraries doing? not sure this is the way forward.

auvipy:
yup, that's why not merging but waiting for your thoughts.

guyzyl:
@auvipy  would it be possible to merge it now?

jpadilla (changes requested):
@guyzyl I think this shouldn't be a thing that can be specified, it should always be UTC according to the spec: https://tools.ietf.org/html/rfc7519

>    NumericDate
      A JSON numeric value representing the number of seconds from
      1970-01-01T00:00:00Z UTC until the specified UTC date/time,
      ignoring leap seconds.  This is equivalent to the IEEE Std 1003.1,
      2013 Edition [POSIX.1] definition "Seconds Since the Epoch", in
      which each day is accounted for by exactly 86400 seconds, other
      than that non-integer values can be represented.  See RFC 3339
      [RFC3339] for details regarding date/times in general and UTC in
      particular.

Also, can this be done without introducing a dependency on pytz?

guyzyl (commented):
The change seems fine to me

guyzyl:
If it shouldn't be specified then it's possible to close this PR + ticket.
Regarding pytz, it's possible to drop it as a dependency, will do that if we decide we want to merge the feature.

guyzyl:
@jpadilla Should I close the PR or fix it?

guyzyl:
@jpadilla Should I fix it (as per you comment), or close the PR?

jpadilla:
@guyzyl sorry for the delay. i don't think this is a change/feature we should introduce.

guyzyl:
@jpadilla no worries, makes total sense.

PR #521, opened 2020-10-01
Update return type for jwt.encode

moomoolive:
It seems to me that the return type for the encode method is actually 'bytes' type and not 'str'. While this doesn't make a huge difference working with python code, it gave me some issues yesterday when sending JWT tokens between a Python (flask) REST API and Javascript (Vue) frontend.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -16,7 +16,7 @@ API Reference
     :param str algorithm: algorithm to sign the token with, e.g. ``"ES256"``
     :param dict headers: additional JWT header fields, e.g. ``dict(kid="my-key-id")``
     :param json.JSONEncoder json_encoder: custom JSON encoder for ``payload`` and ``headers``
-    :rtype: str
+    :rtype: bytes
     :returns: a JSON Web Token
 
 .. function:: decode(jwt, key="", verify=True, algorithms=None, options=None, audience=None, issuer=None, leeway=0, verify_expiration=True)


auvipy (approved):

auvipy (changes requested):
can you share what errors you are facing?

moomoolive:
It was honestly pretty simple, since the 'jwt.encode' method outputs a bytes object and not a string, I ended up sending something that looked like this to my front end application:

b'<the jwt token>'

instead of just the token itself. 

The problem arose when I tried to use the token for authorization (decode it), where it would throw an 'invalid header padding warning' because I was sending the bytes object of the token, instead of the string of the token itself.

Hope that makes sense.

auvipy (approved):

tlinhart:
@moomoolive @auvipy I think this PR should be reverted as it seems that https://github.com/jpadilla/pyjwt/pull/513 actually made `encode` return a string. `PyJWT.encode()` uses its base class' `encode` method which returns string, see https://github.com/jpadilla/pyjwt/blob/96ec8639a544c70f4e427942f109dfe59f0d848c/jwt/api_jws.py#L130.

moomoolive:
@tlinhart although the line of code you referenced above should solve the problem, it doesn't seem to work when I try it in my terminal. (BTW I'm using the latest distribution of library 1.7.1)

image

It only becomes a string when I decode the token with ASCII.

image

tlinhart:
Well, in 1.7.1 it returns bytes, but not in master. The current (master) version returns str and thus this PR (which was merged into master 8 days ago) should be reverted.

rayluo:
@tlinhart was right on this:

> Well, in 1.7.1 it returns bytes, but not in master. The current (master) version returns str and thus this PR (which was merged into master 8 days ago) should be reverted.

And, to document my investigation findings here: even though this PR was mentioned in the v2.0.0 release notes and changelogs, but [that line was indeed reverted subsequently](https://github.com/jpadilla/pyjwt/commit/4265a4168cfc07436e14a16cb9535c8078115020). In other words, `encode(...)` returns a string, not bytes.

PR #532, opened 2020-12-06
Remove support for EOL Python 3.5

jdufresne:
Python 3.5 went end-of-life 2020-09-30. See:
https://devguide.python.org/devcycle/#end-of-life-branches


jpadilla (approved):
thanks!

PR #535, opened 2020-12-15
Update typing syntax and usage for Python 3.6+

jdufresne:
Now that Python 2 is not supported, can move away from type comments to
type annotation 🎉.

The typing module is always available, so remove the guards.

Specify the supported Python in the mypy configuration.

Move other mypy configurations to one place. This way, whether tox is
used or not, the same mypy errors appear.

Distribute and install PEP-561 compliant py.typed file. When PyJWT is a
imported as a library, this tells mypy to use the provided type
annotations rather than going through typeshed. This way, the types are
always up to date when running mypy.

Remove outdated ignores since dropping Python 2.

### DIFF
--- a/jwt/api_jws.py
+++ b/jwt/api_jws.py
@@ -1,6 +1,7 @@
 import binascii
 import json
 from collections.abc import Mapping
+from typing import Dict, List, Optional, Type, Union
 
 from .algorithms import requires_cryptography  # NOQA
 from .algorithms import Algorithm, get_default_algorithms, has_crypto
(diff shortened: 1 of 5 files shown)


auvipy (approved):

jpadilla (commented):
@jdufresne thank you for this! q: is that `jwt/py.typed` file supposed to be empty?

jdufresne:
> is that jwt/py.typed file supposed to be empty?

Yes, that is correct. This file is merely a marker for mypy to decide where to find type annotations. Its contents have no meaning.

For details, see:

- https://mypy.readthedocs.io/en/stable/installed_packages.html#making-pep-561-compatible-packages
- https://www.python.org/dev/peps/pep-0561/#packaging-type-information

And here is the mypy source that interacts with this file, notice it only checks that it exists and doesn't open the file:

https://github.com/python/mypy/blob/05d9fb15dcec8bde4e7184ca387e7d8acea68792/mypy/modulefinder.py#L186-L187

jpadilla (approved):

PR #545, opened 2020-12-17
Update code blocks in docs

jdufresne:
Update Python 2 syntax and drop Python-2-only examples.

Fix all Python console prompts to the full ">>> " for correct
highlighting.

Use "pycon" instead of "python" for interactive Python console session.
The "python" lexer is for Python scripts and doesn't interpret the ">>>"
prompt.

Fix all Python console code blocks to more accurately display what the
user will see. For example, when assigning a value to a variable, it
isn't also echoed.

Fix typo ".. code::" → ".. code-block::".

Use "console" instead of "sh" for shell sessions. The "sh" lexer is for
shell scripts. For a complete list of lexers, see:
https://pygments.org/docs/lexers/

Use blacken-docs to run black on code blocks.

### DIFF
--- a/docs/algorithms.rst
+++ b/docs/algorithms.rst
@@ -39,19 +39,20 @@ Specifying an Algorithm
 You can specify which algorithm you would like to use to sign the JWT
 by using the `algorithm` parameter:
 
-.. code-block:: python
+.. code-block:: pycon
 
-    >>> encoded = jwt.encode({'some': 'payload'}, 'secret', algorithm='HS512')
-    'eyJhbGciOiJIUzUxMiIsInR5cCI6IkpXVCJ9.eyJzb21lIjoicGF5bG9hZCJ9.WTzLzFO079PduJiFIyzrOah54YaM8qoxH9fLMQoQhKtw3_fMGjImIOokijDkXVbyfBqhMo2GCNu4w9v7UXvnpA'
+    >>> encoded = jwt.encode({"some": "payload"}, "secret", algorithm="HS512")
+    >>> print(encoded)
+    eyJhbGciOiJIUzUxMiIsInR5cCI6IkpXVCJ9.eyJzb21lIjoicGF5bG9hZCJ9.WTzLzFO079PduJiFIyzrOah54YaM8qoxH9fLMQoQhKtw3_fMGjImIOokijDkXVbyfBqhMo2GCNu4w9v7UXvnpA
 
 When decoding, you can also specify which algorithms you would like to permit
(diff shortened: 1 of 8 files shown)


jpadilla:
@jdufresne you've outdone yourself. I'm very thankful for all this awesome work you've done. ❤️ 

> Use blacken-docs to run black on code blocks.

It'd be interesting to set this up in CI / pre-commit.

jpadilla (approved):

jdufresne:
> It'd be interesting to set this up in CI / pre-commit.

Good idea. I've included this in the latest revision.

docs/algorithms.rst
```diff
@@ -39,19 +39,20 @@ Specifying an Algorithm
 You can specify which algorithm you would like to use to sign the JWT
 by using the `algorithm` parameter:
 
-.. code-block:: python
+.. code-block:: pycon
```
auvipy, line 42: is it pycon or python?

docs/usage.rst
```diff
@@ -4,26 +4,28 @@ Usage Examples
...
 -------------------------------------
 
-.. code-block:: python
-
-    >>import jwt
-    >>key = 'secret'
-    >>encoded = jwt.encode({'some': 'payload'}, key, algorithm='HS256')
-    'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzb21lIjoicGF5bG9hZCJ9.4twFt5NiznN84AWoo1d7KO1T_yoc0Z6XOpOVswacPZg'
-    >>decoded = jwt.decode(encoded, key, algorithms='HS256')
+.. code-block:: pycon
```
auvipy, line 7: pycon?

auvipy (changes requested):

docs/algorithms.rst
jdufresne: pycon

See here for a complete list of Pygment lexers, which are used by Sphinx: https://pygments.org/docs/lexers/#pygments.lexers.python.PythonConsoleLexer

This code block represents an interactive Python console session. It does not represent a Python script, therefore, pycon is the better choice for code highlighting which understands the `>>> ` prompt.

Same applies to the other review comment.

You can see the same thing in other projects, e.g. Django:
https://github.com/django/django/blob/bb64b99b78a579cb2f6178011a4cf9366e634438/docs/intro/tutorial05.txt#L149-L158

auvipy: didn't know that. thanks for explaining :D

auvipy (approved):

PR #546, opened 2020-12-17
Refactor jwt/jwks_client.py without requests dependency

jdufresne:
Allows dropping a dependency that isn't very necessary.

The requests library was used for a single line of code. This same code
is just as easily expressible using the stdlib, thus alllows removing a
dependency.

Tests were adjusted to mock this new approach.

### DIFF
--- a/jwt/jwks_client.py
+++ b/jwt/jwks_client.py
@@ -1,27 +1,18 @@
+import json
+import urllib.request
+
 from .api_jwk import PyJWKSet
 from .api_jwt import decode as decode_token
 from .exceptions import PyJWKClientError
 
-try:
-    import requests
-
-    has_requests = True
-except ImportError:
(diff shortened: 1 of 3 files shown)


jwt/jwks_client.py
```diff
@@ -1,27 +1,18 @@
...
-                "Missing dependencies for `PyJWKClient`. Run `pip install pyjwt[jwks-client]` to install dependencies."
-            )
-
         self.uri = uri
 
     def fetch_data(self):
-        r = requests.get(self.uri)
-        return r.json()
+        with urllib.request.urlopen(self.uri) as response:
+            return json.load(response)
```
jpadilla, line 15: 🙈 heh this was so much easier than all the cruft I had added.

jpadilla (approved):

PR #549, opened 2020-12-17
Replace int_from_bytes() with builtin int.from_bytes()

jdufresne:
Follows upstream cryptography commit:
https://github.com/pyca/cryptography/commit/5528a3182fdd6ed1c44c126d451a87bcf39e79de

Since Python 3.2, this bytes to an int is a native feature.

### DIFF
--- a/jwt/algorithms.py
+++ b/jwt/algorithms.py
@@ -42,7 +42,6 @@ try:
         load_pem_public_key,
         load_ssh_public_key,
     )
-    from cryptography.utils import int_from_bytes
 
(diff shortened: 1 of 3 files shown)


jpadilla:
@jdufresne do we need to ensure any specific cryptography version?

jdufresne:
Nope.

The `int.from_bytes()` method is built into Python since 3.2. For docs, see: https://docs.python.org/3/library/stdtypes.html#int.from_bytes

jpadilla (approved):

PR #550, opened 2020-12-17
Add PyPy3 to the test matrix

jdufresne:


jdufresne:
Looks like this fails on Windows. I'll see what options are available.

PR #555, opened 2020-12-17
Replace merge_dict() with builtin dict unpacking generalizations

jdufresne:
Merging two dict is support since Python 3.5 using the ** syntax.

See:

- https://docs.python.org/3.9/whatsnew/3.5.html#whatsnew-pep-448
- https://www.python.org/dev/peps/pep-0448/


jpadilla (approved):
til, nice!

PR #557, opened 2020-12-18
Do not mutate the input payload in PyJWT.encode()

jdufresne:


Dreamsorcerer:
Looks good

jwt/api_jwt.py
```diff
@@ -47,6 +47,7 @@ def encode(
             )
 
         # Payload
+        payload = payload.copy()
```
jpadilla, line 50: good catch!

jpadilla (approved):

PR #564, opened 2020-12-19
Type hint jwt.utils module

jdufresne:


jwt/utils.py
```diff
@@ -1,17 +1,19 @@
 import base64
 import binascii
 import struct
+from typing import Union
 
 try:
+    from cryptography.hazmat.primitives.asymmetric.ec import EllipticCurve
     from cryptography.hazmat.primitives.asymmetric.utils import (
         decode_dss_signature,
         encode_dss_signature,
```
Dreamsorcerer, line 10: Slightly off-topic, but there should probably be some kind of runtime check that the above functions were successfully imported (if only to provide better user feedback). See, for example; https://github.com/aio-libs/aiohttp-security/blob/master/aiohttp_security/session_identity.py#L21

jwt/utils.py
```diff
@@ -1,17 +1,19 @@
...
     )
 except ImportError:
-    pass
+    EllipticCurve = object  # type: ignore
```
Dreamsorcerer, line 13: `Any` is probably a better choice here.
`Any` basically means dynamic typing, i.e. we don't know if it will work or not.
`object` basically means we don't care what type is passed, it is always type safe.
If you defined this without the import above, so Mypy used `object`, you would find there are typing errors in those functions (object has not attribute key_size).

Alternatively, you could even define it with:
```
class EllipticCurve(Protocol):
    @property
    def key_size(self) -> int: ...
```

Probably not worth it though, given that these functions will fail anyway if the import failed.

Dreamsorcerer (changes requested):

jwt/utils.py
jdufresne: Thanks switched to `Any`.

mypy runs with cryptograpy installed, so the real class will always exist during mypy runs. IMO, we should use the upstream definition as much as possible rather than duplicating it.

jwt/utils.py
jdufresne: That already exists here. Please feel free to submit PRs to improve the user facing messaging.


https://github.com/jpadilla/pyjwt/blob/5bf8e70f26309b0a5ceda49ce0a7e97e89eee764/jwt/api_jws.py#L117-L120

Dreamsorcerer: Right, I suppose these function are for internal use, so not relevant.

Dreamsorcerer (approved):

jpadilla (approved):

PR #566, opened 2020-12-20
Fix tox "manifest" environment to pass

jdufresne:
The command `tox -e manifest` now passes.

Added missing files to the sdist:

- .pre-commit-config.yaml
- CODE_OF_CONDUCT.md
- jwt/py.typed
- docs/

Alphabetized MANIFEST.in

Moved .flake8 to setup.cfg to avoid the need to include yet another file
to MANIFEST.in.

Exclude codecov.yml. This is for CI only.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,0 @@
-[flake8]
-max-line-length = 79
-extend-ignore = E203, E501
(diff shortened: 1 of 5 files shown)


jdufresne:
check-manifest is now handled via pre-commit. This ensures all commits will not include an `MANIFEST.in` issues and removes the need for a separate tox environment target.

jpadilla (approved):

PR #586, opened 2020-12-23
To make new version backward compatible

jitendra29mishra:
Some project code develops with verify argument with value false.

In the new version, this argument was dropped.


jpadilla:
This was a conscious breaking change in v2 https://github.com/jpadilla/pyjwt/blob/master/CHANGELOG.md#dropped-deprecated-verify-param-in-jwtdecode

Issue #602, opened 2021-01-11
encode()/decode() key parameter type should not be str


WolfgangFellger:
The ``jwt.encode()`` and `jwt.decode()` functions declare parameter `key` to be `str`. This seems to have been the case for a while, but with new release mypy will now complain that
`Argument "key" has incompatible type "bytes"; expected "str"  [arg-type]`
when this parameter is given a bytes object.

Giving it a string does not seem to make sense though, the key is inherently binary (and indeed the first thing HMAC does is throw the key into `force_bytes`: https://github.com/jpadilla/pyjwt/blob/fdfd6871/jwt/algorithms.py#L173-174).

Everything works as intended when you pass in `bytes`, it's just the type declaration.

## Expected Result

Parameter should accept `bytes` (and possibly _not_ accept `str`)

## Actual Result

mypy reports a type error when using bytes.
When using str it is impossible to pass random bits (since it will be run through `.encode('utf-8')`)

## Reproduction Steps

```python
import secrets
import jwt

secret = secrets.token_bytes(32)
token = jwt.encode(payload={"aud": "foobar"}, key=secret, algorithm="HS256")
result = jwt.decode(
    token,
    key=secret,
    audience="foobar",
    verify=True,
    algorithms=["HS256"],
    options={"require": ["aud"]},
)
```

Runs fine, but mypy will report errors.

asthasr:
It should also accept the private key types from Cryptography (on encode) and the public key types (on decode).

Example:

```python
import jwt
from cryptography.hazmat.primitives.asymmetric import rsa

private_key = rsa.generate_private_key(public_exponent=65537, key_size=4096)
public_key = private_key.public_key()
encoded = jwt.encode({"foo": "bar"}, key=private_key, algorithm="RS256")
decoded = jwt.decode(encoded, key=public_key, algorithms=["RS256"])
```

WolfgangFellger:
Yeah, was [discussed over in the PR](https://github.com/jpadilla/pyjwt/pull/605#issuecomment-762779641). Seems like the only practical solution at this time is going to `Any`.

prryplatypus:
Hi there! I know this issue is kinda old, but thought I'd chip in to say that the `jwt.api_jwt.decode_complete` method would probably also need this fixed \:)

ps-george:
This still seems to be an issue

Viicos:
[...]

PR #605, opened 2021-01-12
Fix type annotation for `decode()`/`encode()`'s parameter key

michael-k:
See #602 for details.

The issue suggests that the parameter `key` should not accept `str`.  I've not included this for backwards compatibility and because I'm not sure if that's really necessary.

Closes jpadilla/pyjwt#602

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,7 +19,7 @@ Fixed
 - Make `kty` mandatory in JWK to be compliant with RFC7517. `#624 <https://github.com/jpadilla/pyjwt/pull/624>`__
 - Allow JWK without `alg` to be compliant with RFC7517. `#624 <https://github.com/jpadilla/pyjwt/pull/624>`__
 - ``encode()``, ``decode()``, and ``decode_complete()``'s parameter key now
-  accepts and defaults to `bytes` `#605 <https://github.com/jpadilla/pyjwt/pull/605>`__
+  accepts ``typing.Any`` and defaults to `bytes` `#605 <https://github.com/jpadilla/pyjwt/pull/605>`__
 
 Added
 ~~~~~
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,8 @@ Fixed
 - Remove padding from JWK test data. `#628 <https://github.com/jpadilla/pyjwt/pull/628>`__
 - Make `kty` mandatory in JWK to be compliant with RFC7517. `#624 <https://github.com/jpadilla/pyjwt/pull/624>`__
 - Allow JWK without `alg` to be compliant with RFC7517. `#624 <https://github.com/jpadilla/pyjwt/pull/624>`__
+- ``encode()``, ``decode()``, and ``decode_complete()``'s parameter key now
+  accepts and defaults to `bytes` `#605 <https://github.com/jpadilla/pyjwt/pull/605>`__
 
 Added
 ~~~~~
--- a/jwt/algorithms.py
+++ b/jwt/algorithms.py
@@ -144,7 +144,7 @@ class NoneAlgorithm(Algorithm):
     """
 
     def prepare_key(self, key):
-        if key == "":
+        if key in ("", b""):
             key = None
 
         if key is not None:
--- a/jwt/algorithms.py
+++ b/jwt/algorithms.py
@@ -144,7 +144,7 @@ class NoneAlgorithm(Algorithm):
     """
 
     def prepare_key(self, key):
-        if key in ("", b""):
+        if not key:
             key = None
 
         if key is not None:
--- a/jwt/api_jws.py
+++ b/jwt/api_jws.py
@@ -1,7 +1,7 @@
 import binascii
 import json
 from collections.abc import Mapping
-from typing import Any, Dict, List, Optional, Type
+from typing import Any, Dict, List, Optional, Type, Union
 
 from .algorithms import (
     Algorithm,
@@ -76,7 +76,7 @@ class PyJWS:
     def encode(
         self,
         payload: bytes,
-        key: str,
+        key: Union[str, bytes],
         algorithm: str = "HS256",
         headers: Optional[Dict] = None,
         json_encoder: Optional[Type[json.JSONEncoder]] = None,
@@ -128,7 +128,7 @@ class PyJWS:
     def decode_complete(
         self,
         jwt: str,
-        key: str = "",
+        key: Union[str, bytes] = b"",
         algorithms: List[str] = None,
         options: Dict = None,
         **kwargs,
@@ -157,7 +157,7 @@ class PyJWS:
     def decode(
         self,
         jwt: str,
-        key: str = "",
+        key: Union[str, bytes] = b"",
         algorithms: List[str] = None,
         options: Dict = None,
         **kwargs,
--- a/jwt/api_jws.py
+++ b/jwt/api_jws.py
@@ -1,7 +1,7 @@
 import binascii
 import json
 from collections.abc import Mapping
-from typing import Any, Dict, List, Optional, Type, Union
+from typing import Any, Dict, List, Optional, Type
 
 from .algorithms import (
     Algorithm,
@@ -76,7 +76,7 @@ class PyJWS:
     def encode(
         self,
         payload: bytes,
-        key: Union[str, bytes],
+        key: Any,
         algorithm: str = "HS256",
         headers: Optional[Dict] = None,
         json_encoder: Optional[Type[json.JSONEncoder]] = None,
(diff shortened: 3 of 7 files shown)


jpadilla:
@jdufresne since you worked in some of the typing stuff recently, thoughts on this change?

jwt/api_jws.py
```diff
@@ -128,7 +128,7 @@ def encode(
     def decode_complete(
         self,
         jwt: str,
-        key: str = "",
+        key: Union[str, bytes] = "",
```
jdufresne, line 131: I think if bytes is the preferred type, this should default to that. Same for the functions.

```suggestion
        key: Union[str, bytes] = b"",
```

CHANGELOG.rst
```diff
@@ -16,6 +16,8 @@ Changed
 Fixed
 ~~~~~
 
+- Fix type annotation for `decode()`/`encode()`'s parameter key `#605 <https://github.com/jpadilla/pyjwt/pull/605>`__
```
jdufresne, line 19: I think this could be more descriptive about what the change is. "Fix"  doesn't really describe what the user should exepct.

jdufresne (commented):
Are there existing tests that cover type bytes? If not, they should be added.

jwt/algorithms.py
```diff
@@ -142,7 +142,7 @@ class NoneAlgorithm(Algorithm):
     """
 
     def prepare_key(self, key):
-        if key == "":
+        if key in ("", b""):
```
jdufresne, line 145: When Python is executed with `-b`, this will emit a `BytesWarning` as it is comparing str to bytes.

michael-k: Does this `BytesWarning` have a deeper meaning or was is created with the transition from Python 2 to 3 in mind?

Should we generalize it to the following or does that have any relevant downsides?
```suggestion
        if not key:
```

Or this?
```suggestion
        if isinstance(key, (str, bytes)) and not key:
```

WolfgangFellger:
Please note that there are [even more types](https://github.com/jpadilla/pyjwt/blob/3993ce1d3503b58cf74699a89ba9e5c18ef9b556/jwt/algorithms.py#L21-L33) that some of the algorithms accept for the `key` argument.

michael-k:
> Please note that there are even more types that some of the algorithms accept for the key argument.

I have no idea how to properly use them in the type annotation.  Everything I've tried so far resulted some error by mypy.

This should work but would probably result in `Union[Any, …]` for every setup that doesn't use `cryptography`:

```python
if MYPY:
    from cryptography.hazmat.primitives.asymmetric.ec import EllipticCurvePrivateKey

def prepare(key: Union[str, bytes, "EllipticCurvePrivateKey"]):
    pass

**Comment by WolfgangFellger** [comment]
Personally I would be okay with just reverting to Any.

**Comment by terencehonles** [comment]
> > Please note that there are even more types that some of the algorithms accept for the key argument.
> 
> I have no idea how to properly use them in the type annotations. Everything I've tried so far resulted some error by mypy.
> 
> This should work but would probably result in `Union[Any, …]` for every setup that doesn't use `cryptography`:
> 
> ```python
> if MYPY:
>     from cryptography.hazmat.primitives.asymmetric.ec import EllipticCurvePrivateKey
> 
> def prepare(key: Union[str, bytes, "EllipticCurvePrivateKey"]):
>     pass
> ```

I'm just checking to see if this relates to a change I was going to make with decode not accepting bytes, but the quoted Python should be expressed

```python
# so KEY_TYPES is not needed at runtime and doesn't break because it's in the TYPE_CHECKING block.
# The constant could be define outside of the block and this future import would not be needed
from __future__ import annotations

from typing import TYPE_CHECKING

if TYPE_CHECKING:
    try:
        from cryptography.hazmat.primitives.asymmetric.ec import EllipticCurvePrivateKey
        KEY_TYPES = Union[str, bytes, EllipticCurvePrivateKey]
    except ImportError:
        KEY_TYPES = Union[str, bytes]

def prepare(key: KEY_TYPES): ...
```

If you need this to be dynamic you could do the following:

```python
_KEY_TYPES = [str, bytes]

# import 1
_KEY_TYPES.append(...)
# import 2
_KEY_TYPES.append(...)

KEY_TYPES = Union[tuple(_KEY_TYPES)]
```

This works because `obj[one, two]` is just passing `(one, two)` to `obj.__getitem__` and as long as the type is a tuple not a list it will be unioned properly.

terencehonles:
It looks like this change is adjacent, but not the change I was looking to make. `jwt` should also be `bytes | str` for the decode functions and there may be other places it's been made too specific and only accepts `str`.

michael-k:
Your first example leads to `Cannot assign multiple types to name "KeyType" without an explicit "Type[...]" annotation`.


`KEY_TYPES = Union[tuple(_KEY_TYPES)]` → `Invalid type alias: expression is not a valid type`.


Also I don't think that something like `_KEY_TYPES.append(…)` could work at all as mypy doesn't import and/or run the code afaik.  The same should be true for try-except.

terencehonles:
Hmm good point. I hadn't completely thought about that and I was going off what I did (awhile ago) **without** testing mypy. At this point I've changed the original type declaration and am playing with getting mypy running so I mainly came here for https://github.com/jpadilla/pyjwt/pull/605#issuecomment-772082918 but if you're not able to import or type against a concrete type I'm assuming you'd need to type with a [protocol](https://docs.python.org/3/library/typing.html#typing.Protocol), but that only comes in Python 3.8.

terencehonles:
It looks like it's on typeshed but it _might_ make sense to declare a common base for private keys https://github.com/python/typeshed/blob/master/stubs/cryptography/cryptography/hazmat/primitives/asymmetric/ec.pyi so that way protocols could be used before Python 3.8. It might be possible to type with Protocols keeping everything behind `if typing.TYPE_CHECKING` and using `from __future__ import annotations` to make sure that any use of Protocols is by `mypy` on Python 3.8 or greater.

terencehonles:
It looks like `Protocol` would be supported pre Python 3.8 via the `typing_extensions`, but not sure if declaring the protocol would be worth it vs just typing with `Any`

michael-k:
@jdufresne any suggestions on how to proceed?

terencehonles:
I believe the consensus here would be to use `Any`. I know you're looking for input from @jdufresne / @jpadilla 's input, but looking at the call chain a call to `j*.decode` it will call `j*.decode_complete` which will likely use the key for:

https://github.com/jpadilla/pyjwt/blob/587997eec3c9ce35d17054d10aa5f2c5403ade4c/jwt/api_jws.py#L217-L224 which is untyped (`Any`)

Looking at the API closer it looks like algorithms are puggable here [pyjwt.jwt.algorithms](https://github.com/jpadilla/pyjwt/blob/master/jwt/algorithms.py) which means the API cannot be type safe. A algorithm will `alg.prepare_key(self, key: T) -> U` and `alg.verify(self, msg: ?, key: U, sig: ?) -> bool` and there's not a good way to type both  of those.

dajiaji:
@michael-k @jpadilla  I think this PR should be changed to use `Any`, and  moved forward.

At present, even the simple sample codes in [the specification](https://pyjwt.readthedocs.io/en/stable/usage.html#encoding-decoding-tokens-with-rs256-rsa) gives mypy warnings.

jwt/algorithms.py
dajiaji: @michael-k I think `if not key` is better choice because it's simple and can cover wider cases.

michael-k:
I've changed the type annotation to `Any`.  @jdufresne could you review this PR again?

jwt/api_jwt.py
```diff
@@ -37,7 +37,7 @@ def _get_default_options() -> Dict[str, Union[bool, List[str]]]:
     def encode(
         self,
         payload: Dict[str, Any],
-        key: str,
+        key: Any,
```
dajiaji, line 40:
```suggestion
        key: Any = b"",
```

jwt/api_jws.py
```diff
@@ -76,7 +76,7 @@ def get_algorithms(self):
     def encode(
         self,
         payload: bytes,
-        key: str,
+        key: Any,
```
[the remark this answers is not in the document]
dajiaji: The omission of the key does not result in signing with the empty string as the private key, and from the perspective of consistency of API design, my suggestion has some validity.  But, anyway, I understood. You don't need to fix it.

dajiaji:
@michael-k One last thing: I think you should update https://github.com/jpadilla/pyjwt/blob/master/docs/api.rst.

auvipy (changes requested):
also, please rebase

ps-george:
bump - this is still an issue

michael-k:
I've rebased, but I don't seem to be able to re-open this PR.

lcy0321:
Hi @michael-k, @auvipy,
I think this is still an issue, will this PR be re-opened?

PR #617, opened 2021-02-16
Ignore coverage files generated during test runs

makusu2:
During test runs, new coverage files are generated, such as `.coverage.hostname.number.number`. They get deleted at the end of `tox`, but in the interest of not accidentally committing them during a `tox` run, they should be ignored.


makusu2:
@jpadilla there's not really any rush on this, but it should be a pretty easy PR. Bump

jpadilla (approved):

PR #620, opened 2021-02-27
Add missing exceptions.InvalidKeyError to jwt module __init__ imports.

aarcamp:
All other error classes are already here, so I'm assuming this was an oversight.


jpadilla (approved):
@aarcamp thanks! mind updating the CHANGELOG file as well?

PR #627, opened 2021-03-09
Add support for multiple issuers

### LINKED ISSUE #433: Allow multiple issuers?
Does it seem reasonable to allow PyJWT to accept a list of possible `iss`uer arguments, in the same way that `aud` does?  (It's possible I'm just lazy and trying to avoid try-catch blocks :joy:)

patrick91:
Closes #433

This PR adds support for specifing multiple issuers, similar to what we do for audience.

### DIFF
--- a/jwt/api_jwt.py
+++ b/jwt/api_jwt.py
@@ -212,7 +212,11 @@ class PyJWT:
         if "iss" not in payload:
             raise MissingRequiredClaimError("iss")
 
-        if payload["iss"] != issuer:
+
+        if isinstance(issuer, str):
+            issuer = [issuer]
+
+        if payload["iss"] not in issuer:
             raise InvalidIssuerError("Invalid issuer")
 
 
--- a/jwt/api_jwt.py
+++ b/jwt/api_jwt.py
@@ -212,7 +212,6 @@ class PyJWT:
         if "iss" not in payload:
             raise MissingRequiredClaimError("iss")
 
-
         if isinstance(issuer, str):
             issuer = [issuer]
 
--- a/tests/test_api_jwt.py
+++ b/tests/test_api_jwt.py
@@ -419,6 +419,12 @@ class TestJWT:
         token = jwt.encode(payload, "secret")
         jwt.decode(token, "secret", issuer=issuer, algorithms=["HS256"])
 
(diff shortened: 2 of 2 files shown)


jwt/api_jwt.py
```diff
@@ -212,7 +212,10 @@ def _validate_iss(self, payload, issuer):
         if "iss" not in payload:
             raise MissingRequiredClaimError("iss")
 
-        if payload["iss"] != issuer:
+        if isinstance(issuer, str):
```
jpadilla, line 215: do we also need to guard against other invalid values for `issuer`?

see https://github.com/jpadilla/pyjwt/blob/master/jwt/api_jwt.py#L128-L129

jpadilla (changes requested):

tests/test_api_jwt.py
```diff
@@ -419,6 +419,12 @@ def test_check_issuer_when_valid(self, jwt):
         token = jwt.encode(payload, "secret")
         jwt.decode(token, "secret", issuer=issuer, algorithms=["HS256"])
 
+    def test_check_multiple_issuers(self, jwt):
+        issuer = "urn:foo"
+        payload = {"some": "payload", "iss": "urn:foo"}
+        token = jwt.encode(payload, "secret")
+        jwt.decode(token, "secret", issuer=[issuer, "urn:bar"], algorithms=["HS256"])
```
dimaqq, line 426: The spec insists on issuers being HTTPS URLs:
https://openid.net/specs/openid-connect-core-1_0.html#Terminology

```suggestion
        jwt.decode(token, "secret", issuer=[issuer, "https://other.test"], algorithms=["HS256"])
```

dimaqq: (arguably `pyjwt` could be used in other contexts, not just oidc)

peterfarrell:
I'm interested in this feature.

ChegeBryan:
> I'm interested in this feature.

That makes two of us.

ChegeBryan:
@jpadilla is support for multiple issuers coming soon?

jpadilla:
there were some unresolved comments on this pull request, I'd gladly review again another.

PR #628, opened 2021-03-13
Remove padding from JWK test data.

dajiaji:


jpadilla (commented):
Any specific reason this is required?

dajiaji:
Thanks for checking this PR.

base64url encoding can skip padding("="). Many jwk implementations that I know of omit the padding, and I thought that test data would be better suited to the common implementation.

However, skipping padding is not mandatory so I'll leave it up to you to decide if you want to accept this PR.

jpadilla (approved):
Thanks, can you add this to the CHANGELOG?

PR #633, opened 2021-03-20
Use UTC time for exp/nbf validation.

### LINKED ISSUE #487: validate nbf uses local timezone for its current time. I believe it should use UTC

### LINKED ISSUE #631: UTC time for exp claim works as not expected
Hello,
better then my pure English is a simple pytest snippet:
```
# Python 3.9.1

from datetime import datetime

import jwt  # pyjwt == 2.0.0
import pytest


def test_encode_and_decode():
    JWT_PRIVATE_KEY = 'bibalabibaloca'
    JWT_ALGORITHM = 'HS256'
    now = datetime.now().timestamp()
    utcnow = datetime.utcnow().timestamp()
    delta = 7 * 60  # 7 minutes

    # if someone did not know - True for these, who lives in Russia
    assert int(now / 10) > int(utcnow / 10)

    # UTC case
    exp = int(utcnow + delta)
    iat = int(utcnow)

    payload_in = {
        'exp': exp,
        'iat': iat
    }
    token = jwt.encode(
        payload_in,
        JWT_PRIVATE_KEY,
        JWT_ALGORITHM
    )

    with pytest.raises(jwt.exceptions.ExpiredSignatureError):
        payload_out = jwt.decode(
            jwt=token,
            leeway=0,
            key=JWT_PRIVATE_KEY,
            algorithms=[JWT_ALGORITHM],
        )

    # local time case
    exp = int(now + delta)
    iat = int(now)

    payload_in = {
        'exp': exp,
        'iat': iat
    }
    token = jwt.encode(
        payload_in,
        JWT_PRIVATE_KEY,
        JWT_ALGORITHM
    )

    payload_out = jwt.decode(
        jwt=token,
        leeway=0,
        key=JWT_PRIVATE_KEY,
        algorithms=[JWT_ALGORITHM],
    )

    assert payload_in == payload_out
```

That test has passed for me. Documentation says: you have to use UTC time for claims, as `jwt.decode` uses UTC time to compare such claims like `exp` . From the test case we see, that this statement is False, or I missed something.
Thanks.

dajiaji:
To be compliant with RFC7519, Update exp/nbf validation to use UTC time instead of local time.

Close #487 
Close #631

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,7 @@ Fixed
 ~~~~~
 
 - Remove padding from JWK test data. `#628 <https://github.com/jpadilla/pyjwt/pull/628>`__
+- Use UTC time for exp/nbf validation. `#633 <https://github.com/jpadilla/pyjwt/pull/633>`__
 
 Added
 ~~~~~
--- a/jwt/api_jwt.py
+++ b/jwt/api_jwt.py
@@ -1,5 +1,5 @@
 import json
-from calendar import timegm
+import time
 from collections.abc import Iterable, Mapping
 from datetime import datetime, timedelta
 from typing import Any, Dict, List, Optional, Type, Union
@@ -54,7 +54,9 @@ class PyJWT:
         for time_claim in ["exp", "iat", "nbf"]:
             # Convert datetime to a intDate value in known time-format claims
             if isinstance(payload.get(time_claim), datetime):
-                payload[time_claim] = timegm(payload[time_claim].utctimetuple())
+                payload[time_claim] = int(
+                    time.mktime(payload[time_claim].utctimetuple())
+                )
 
         json_payload = json.dumps(
             payload, separators=(",", ":"), cls=json_encoder
@@ -130,7 +132,7 @@ class PyJWT:
 
         self._validate_required_claims(payload, options)
 
-        now = timegm(datetime.utcnow().utctimetuple())
+        now = datetime.utcnow().timestamp()
 
         if "iat" in payload and options["verify_iat"]:
             self._validate_iat(payload, now, leeway)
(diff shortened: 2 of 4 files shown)


auvipy (approved):

jpadilla (commented):
@dajiaji thanks for working on this. For posterity's sake, could you walk us through why the current _thing_ is wrong, and how your changes fix it?

dajiaji:
@jpadilla @auvipy Sorry for bothering you on this PR. 

I misunderstood datetime() related spec and behavior. Current implementation has no problem and does not need to be updated. I'll withdraw this PR.

PR #645, opened 2021-04-09
Allow to verify with private key on ECAlgorithm, as well as on Ed25519Algorithm.

dajiaji:
At the current implementation of Ed25519Algorithm, verify() can accept a private key as a verification key (See https://github.com/jpadilla/pyjwt/blob/master/jwt/algorithms.py#L583-L584). However, ECAlgorithm's verify() cannot.

I think the both of two should work in the same way. There are two options:
1. Remove private key support from Ed25519Algorithm's verify().
2. Add private key support to ECAlgorithm's verify().

Considering backward-compat, I selected 2.

### DIFF
--- a/jwt/algorithms.py
+++ b/jwt/algorithms.py
@@ -427,6 +427,8 @@ if has_crypto:
                 return False
 
             try:
+                if isinstance(key, EllipticCurvePrivateKey):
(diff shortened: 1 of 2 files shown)


jpadilla (approved):
thanks!

PR #647, opened 2021-04-16
Clearly warn against computing algorithms from the token’s alg

andersk:
This seems to be a common temptation (python-social-auth/social-core#514, https://github.com/twilio/twilio-python/pull/560#discussion_r614487375), so we should clarify why it’s a bad idea.


jpadilla (approved):
Yikes, good find! 👍

PR #654, opened 2021-04-29
Complete `jwt` documentation

johachi:
## Description

- Added documentation for `jwt.decode_complete`.
  I'm honestly unsure if everything needs to be duplicated or not since the only thing that is different to `jwt.decode` is the return value. I'm happy to just write `Parameters: Sames as jwt.decode` if that is preferred.
- I previously wrote (e.g.) `verify_aud=True`` but will be ignored if ``verify_signature`` is ``False``.` for all the options parameters except `require` and `verify_signature`, but that is a bad way of describing it. 

  What I wanted to write was that the value is `True` by default, but will default to `False` if `verify_signature` is set to false. In other words, they default to the same value as `verify_signature`. I, therefore, think that a better way of writing this is `verify_aud=verify_signature`, and then move `verify_signature=True` to the top. This way, it is understandable that it is possible to set (e.g.) `verify_aud=True` and `verify_signature=False`.
- Added a warning regarding `exp`, `iat` and `nbf` only being verified if present.

**Screenshot (Firefox Developers Edition 89.0b5)**
Documentation

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -69,7 +69,9 @@ API Reference
 
 .. function:: decode_complete(jwt, key="", algorithms=None, options=None, audience=None, issuer=None, leeway=0)
 
-    Verify the ``jwt`` token signature and return the token claims.
+    Identical to ``jwt.decode`` except for return value which is a dictionary containing the token header (JOSE Header),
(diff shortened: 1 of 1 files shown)


auvipy (approved):

jpadilla (approved):
✨ thanks!

PR #662, opened 2021-05-29
Add a type declaration for PyJWK.key

coderanger:


coderanger:
In theory this can probably be deduced but the generic-y indirection confused Pyright so just hardwiring it for simplicity and reference.

coderanger:
Any interest in this?

auvipy (commented):
what is the value of it?

coderanger:
Anyone using a type checker will not need to cast/mark this and a tiny QoL boost for anyone using an IDE that uses type data for autocomplete.

PR #666, opened 2021-06-19
support decoding and encoding DEFLATE-encoded JWT bodies

Roguelazer:
These are used in the [Smart Health Card](https://spec.smarthealth.cards/) framework that several regions (including California and Quebec) are using for COVID-19 proofs of vaccination, and are supported by some other JWT libraries, such as [jwt-go](https://github.com/dgrijalva/jwt-go/pull/102).

If y'all like this approach, I'm glad to write tests. I could also see putting this logic in `api_jws.py`.

### DIFF
--- a/jwt/api_jwt.py
+++ b/jwt/api_jwt.py
@@ -3,6 +3,7 @@ from calendar import timegm
 from collections.abc import Iterable, Mapping
 from datetime import datetime, timedelta
 from typing import Any, Dict, List, Optional, Type, Union
+import zlib
 
 from . import api_jws
 from .exceptions import (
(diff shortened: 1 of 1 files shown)


tahnok:
I'd also volunteer to help write some tests. I am also looking at decoding the data in an SHC encoded JWT and at the moment pyjwt cannot be used for this

auvipy:
> I'd also volunteer to help write some tests. I am also looking at decoding the data in an SHC encoded JWT and at the moment pyjwt cannot be used for this
> 
> Overall, change looks solid and straightforward

you are welcome to do so

akx:
Hmm – I'd be a little wary against adding a non-spec feature like this. In fact, the golang library PR @Roguelazer merged was actually _not_ merged due to the same concern: https://github.com/dgrijalva/jwt-go/pull/102#issuecomment-209151496

I think the better approach would be to refactor payload decoding (lines 98..104 in the preimage) to a separate function that can be easily augmented in a subclass.

jpadilla:
Closing this one for now, same reason as #753.

Issue #626, opened 2021-03-08
PyJWKClient raises error when key omits "use" (Public Key Use) parameter


mjr9804:
`PyJWKClient.get_signing_keys` is hard-coded to look for keys that have a `use` (Public Key Use) value of `sig`. This causes errors when working with JWKs that omit this parameter. I believe this is an optional parameter according to [RFC 7517, Section 4.2](https://tools.ietf.org/html/rfc7517#section-4.2).

Adjusting the check at https://github.com/jpadilla/pyjwt/blob/master/jwt/jwks_client.py#L32 fixes the problem, but I'm not sure what the best solution is. It seems you could either assume that a key that omits `use` is a signing key, ignore any keys with `"use": "enc"`, or maybe expose some configuration knob.

## Expected Result

`PyJWKClient` should be able to parse signing keys from JWKs that omit the optional `use` parameter.

## Actual Result

A `PyJWKClientError` is raised.

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../Python/3.7/lib/python/site-packages/jwt/jwks_client.py", line 31, in get_signing_keys
    raise PyJWKClientError("The JWKS endpoint did not contain any signing keys")
jwt.exceptions.PyJWKClientError: The JWKS endpoint did not contain any signing keys
```

## Reproduction Steps

```python
import jwt

# Sample JWK that omits 'use'
SAMPLE_URL = 'https://gist.githubusercontent.com/mjr9804/83e8388cc5d2a03e54938da189668770/raw/5514eec3c12714eb2a57bf180810b23ac841ad9b/jwk.json'
jwks_client = jwt.PyJWKClient(SAMPLE_URL)
jwks_client.get_signing_keys()
```

## System Information

    $ python -m jwt.help

```
{
    "cryptography": {
      "version": "3.2"
    },
    "implementation": {
      "name": "CPython",
      "version": "3.7.3"
    },
    "platform": {
      "release": "19.6.0",
      "system": "Darwin"
    },
    "pyjwt": {
      "version": "2.0.1"
    }
}
```

PR #668, opened 2021-07-06
Assume JWK is valid for signing if "use" is omitted

Klavionik:
There's an issue https://github.com/jpadilla/pyjwt/issues/626 about this problem. I ran into it today too.

Currently, when getting signing keys, PyJWKClient assumes that a key is intended for signing, only if its "use" claim is set to "sig". This is not entirely correct, as this claim is optional. If it's omitted, then the key can be used both for signing and encryption. 

I guess the correct way to process fetched keys is to check that their "use" claim is either "sig" or missing. In this case, we can safely append it to the list of signing keys.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ Changed
 
 Fixed
 ~~~~~
+- Assume JWK without the "use" claim is valid for signing as per RFC7517 `#668 <https://github.com/jpadilla/pyjwt/pull/668>`__
 
 Added
 ~~~~~
--- a/jwt/jwks_client.py
+++ b/jwt/jwks_client.py
@@ -29,7 +29,7 @@ class PyJWKClient:
         signing_keys = []
 
         for jwk_set_key in jwk_set.keys:
-            if jwk_set_key.public_key_use == "sig" and jwk_set_key.key_id:
+            if jwk_set_key.public_key_use in ["sig", None] and jwk_set_key.key_id:
                 signing_keys.append(jwk_set_key)
 
         if len(signing_keys) == 0:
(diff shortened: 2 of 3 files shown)


dajiaji:
It's a small fix and I think it should be merged.

yhaojin:
I also ran into this issue today where I was trying to get signing keys from a jwks endpoint without the key "use", and the error `PyJWKClientError("The JWKS endpoint did not contain any signing keys") `was raised.

Refer to rfc7517 section 4.2: 
> The "use" (public key use) parameter identifies the intended use of the public key.  The "use" parameter is employed to indicate whether a public key is used for encrypting data or verifying the signature on data.
...
The "use" value is a case-sensitive string.  Use of the "use" member is OPTIONAL, unless the application requires its presence.

One workaround that I did was to overwrite the `get_signing_keys` to remove the requirement. I definitely support this PR to be merged.

jpadilla (approved):
Thanks! I think there might still be some cases we're missing. If you're up to it, it'd be great to compare notes/behavior against https://github.com/panva/jose/blob/main/src/jwks/remote.ts#L100

Klavionik:
As for the "use" claim, that library has a similar behavior - "check if the claim is present in the object, and only then check that it's set to "sig". Otherwise, just ignore this claim during the check pipeline."

Overall, its behavior and checks are more complex, that's for sure. Though we could implement check for the `key_ops` array, as it is a bit similar to "use" and specifies uses for the JWK.

jpadilla (approved):
@Klavionik mind adding a line for this in the CHANGELOG? thanks again

Klavionik:
@jpadilla I've updated the changelog. Realized I messed up with the committer name, had to manipulate the history and force-push the branch. First time contributing and pretty nervous, sorry about that. :)

PR #674, opened 2021-07-23
feat: allow PyJWKClient to receive HTTPS context

pgbezerra:
Sometimes you have a SSO with a customized certificate, or you just need
to ignore certificate validation for testing or development pourposes.
This change allow us to call PyJWKClient with a customized SSLContext,
that allow us to customize JWKS URL request.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,8 @@ Fixed
 Added
 ~~~~~
 
+- Allow PyJWKClient to receive HTTPS context
+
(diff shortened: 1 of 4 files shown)


jpadilla (commented):
@pgbezerra thanks for putting this together!

tbh I think instead I'd prefer users to extend PyJWKClient's fetch_data to do what they want. I still think that'd be a valuable example in the docs if you're up for it.

pgbezerra:
Well, this works too.
I'll get some time to do this.

PR #675, opened 2021-07-25
Add support for Ed448/EdDSA.

dajiaji:
Hi, @jpadilla.

I've added support for Ed448/EdDSA. In order to implement the curve Ed448, I renamed the `Ed25519Algorithm` class to `OKPAlgorithm`. I think it is better because it is more in line with `ECAlgorithm`.

### DIFF
--- a/jwt/algorithms.py
+++ b/jwt/algorithms.py
@@ -22,6 +22,10 @@ try:
         EllipticCurvePrivateKey,
         EllipticCurvePublicKey,
     )
+    from cryptography.hazmat.primitives.asymmetric.ed448 import (
+        Ed448PrivateKey,
+        Ed448PublicKey,
+    )
     from cryptography.hazmat.primitives.asymmetric.ed25519 import (
         Ed25519PrivateKey,
         Ed25519PublicKey,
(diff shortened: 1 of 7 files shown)


jwt/algorithms.py
```diff
@@ -534,9 +538,9 @@ def verify(self, msg, key, sig):
             except InvalidSignature:
                 return False
 
-    class Ed25519Algorithm(Algorithm):
+    class OKPAlgorithm(Algorithm):
```
jpadilla, line 541: I think we might need to consider this a breaking change.

dajiaji: Thanks for your time reviewing this PR.

I think Algorithm classes are not disclosed via jwt/__init__.py so it is not necessary for us to regard this change as a breaking change.

As you can see, there is no change on public class tests (e.g., jwt_test.py, jws_test.py, jwk_test.py) except for [here](https://github.com/jpadilla/pyjwt/pull/675/files#diff-a18f5631d08da144d86418dfbc5b72725aa3801d9eda3d0a38df7acc72e1b8ccR164) where a test assertion checks an instance type of `jwk.Algorithm` returned.

jpadilla (approved):

jwt/algorithms.py
dajiaji: @auvipy Hi, could you review and merge it?

auvipy: my only  concern is  what happen to Ed25519Algorithm? can we use deprecation warning to it and introduce  OKPAlgorithm at the same time?

auvipy (changes requested):

jwt/algorithms.py
dajiaji: > can we use deprecation warning to it and introduce OKPAlgorithm at the same time?

We don't need to use deprecation warning because `Ed25519Algorithm` is a private class and this change does not break backward-compat. All of the tests for public classes and methods are passed successfully.

auvipy (approved):

PR #682, opened 2021-08-19
Implement AWS KMS algorithm

### LINKED ISSUE #681: Support using AWS KMS to sign and verify.
AWS KMS allows for signing and verifying arbitrary binary strings using a variety of algorithms. Create an "algorithm" that wraps the `boto3 kms` methods of `sign` and `verify`. The key would be the key id, key ARN, etc, created in KMS.

This may be slightly difficult as the algorithm would have to handle not only the key but the signing algorithm. If desired, this could be a separate library built on top of PyJWT.

More info:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/kms.html#KMS.Client.verify

JosiahDub:
Hello! This adds a new "algorithm" that wraps the AWS KMS `sign` and `verify` methods. If given a valid ARN (and provided that AWS credentials are properly set) this algorithm allows for encoding and decoding JWTs without ever seeing a private or public key.

Example usage:

```
import jwt
arn = 'arn:aws:kms:us-east-1:2***0:key/***'
jtw_str = jwt.encode({'foo': 'bar'}, arn, algorithm='AWS-KMS')
dct = jwt.decode(jwt_str, arn, algorithms=['AWS-KMS'])
```

This is a WIP as there are some issues to work out. 

- How does the user change the signing algorithm? 
- What if the AWS credentials are not set up? 
- How to mock the API for testing purposes?
Any ideas are welcome.

Closes #681

### DIFF
--- a/jwt/algorithms.py
+++ b/jwt/algorithms.py
@@ -1,6 +1,7 @@
 import hashlib
 import hmac
 import json
+import re
 
 from .exceptions import InvalidKeyError
 from .utils import (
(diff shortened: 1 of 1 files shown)


jpadilla:
Thanks for working on this. I don't think it makes much sense to include something like this in the library per se, might be better as an addon / third party lib.

PR #690, opened 2021-09-29
Chore: inline Variables that immediately Returned

yezz123:
- I just fix some issues related to code.
- Refactor some Parts of the code.

### DIFF
--- a/jwt/api_jws.py
+++ b/jwt/api_jws.py
@@ -90,9 +90,6 @@ class PyJWS:
         if headers and "alg" in headers and headers["alg"]:
             algorithm = headers["alg"]
 
-        if algorithm not in self._valid_algs:
-            pass
-
         # Header
         header = {"typ": self.header_typ, "alg": algorithm}
 
--- a/jwt/algorithms.py
+++ b/jwt/algorithms.py
@@ -243,22 +243,21 @@ if has_crypto:
             self.hash_alg = hash_alg
 
         def prepare_key(self, key):
-            if isinstance(key, RSAPrivateKey) or isinstance(key, RSAPublicKey):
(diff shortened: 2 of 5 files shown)


auvipy (changes requested):
can you please update tests to see this changes wont regress?

yezz123:
> can you please update tests to see this changes wont regress?

It's not a new feature to resolve in the tests, cause this is just a simple way to refactor some parts of code, I didn't add any function that why 😀

auvipy:
I understand but still the changes are significant so you can add some additional test on existing tests only

yezz123:
> I understand but still, the changes are significant so you can add some additional tests on existing tests only

I don't see where we can add this test, cause this is just some simple changes that why 🤔

auvipy:
https://github.com/jpadilla/pyjwt/tree/master/tests tests for related modules can be found here

beliaev-maksim:
@auvipy,
I just review changes. It is purely refactoring, no functionality was changed, you do not need additional tests

beliaev-maksim (approved):
pure refactoring, looks good

dajiaji:
+1

jwt/api_jws.py
```diff
@@ -90,9 +90,6 @@ def encode(
         if headers and "alg" in headers and headers["alg"]:
             algorithm = headers["alg"]
 
-        if algorithm not in self._valid_algs:
```
auvipy, line 93: may I know why this was removed?

auvipy (approved):

jwt/api_jws.py
beliaev-maksim: Because it just `pass` statement, it has no influence except time spent for if condition check

PR #695, opened 2021-10-05
refactor: Add type information for options dictionary

kkirsche:
This pull request adds support for a more strictly typed options dictionary to support individuals to self-discover the available options without the documentation or evaluating the source code.


kkirsche:
Ah, oops this is Python 3.8 and above :( sorry

Issue #718, opened 2021-12-26
The result of encoding is different from the official website（https://jwt.io/）.


miloira:
The official website result is `eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6MSwibmFtZSI6ImFkbWluIiwiZXhwIjoxNTE2MjM5MDIyfQ.Wmw6IKBTkRK0c7qX1E2rEGb8D6cqjnC6l3FFOT4WvOs`

## Expected Result

eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6MSwibmFtZSI6ImFkbWluIiwiZXhwIjoxNTE2MjM5MDIyfQ.Wmw6IKBTkRK0c7qX1E2rEGb8D6cqjnC6l3FFOT4WvOs
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6MSwibmFtZSI6ImFkbWluIiwiZXhwIjoxNTE2MjM5MDIyfQ.Wmw6IKBTkRK0c7qX1E2rEGb8D6cqjnC6l3FFOT4WvOs

## Actual Result

eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6MSwibmFtZSI6ImFkbWluIiwiZXhwIjoxNTE2MjM5MDIyfQ.Wmw6IKBTkRK0c7qX1E2rEGb8D6cqjnC6l3FFOT4WvOs
eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpZCI6MSwibmFtZSI6ImFkbWluIiwiZXhwIjoxNTE2MjM5MDIyfQ.JnHKRoQGqK0YgsCft_xowLNQkWOdeJrMeR3I10tVBs4

## Reproduction Steps

```python
import json
import base64
import hmac
import hashlib
import jwt

headers = {"alg":"HS256","typ":"JWT"}

payload = {"id":1,"name":"admin","exp":1516239022}

def jwt_encode(header, payload, key):
    json_header = json.dumps(header,separators=(",", ":"))
    json_payload = json.dumps(payload,separators=(",", ":"))

    header_b64 = base64.urlsafe_b64encode(json_header.encode()).replace(b'=', b'')
    payload_b64 = base64.urlsafe_b64encode(json_payload.encode()).replace(b'=', b'')

[...]

PR #719, opened 2021-12-26
The result of jwt.encode is the same as https://jwt.io/ #718

miloira:


auvipy (approved):

jpadilla:
Apart from the obvious of matching output to jwt.io, is there any real need or benefit from this?

sabrina981:
I don’t know, I only know I am being bullied and harassed for over a year

miloira:
> Apart from the obvious of matching output to jwt.io, is there any real need or benefit from this?


At least making pyjwt seems right.

PR #723, opened 2022-01-16
Add detached payload support for JWS encoding and decoding

fviard:
Specifications allow to have JWS with unencoded detached payloads.
This changeset adds detached payload support for encoding and decoding
functions.

For encoding, detached payload can be enabled by setting the
"is_payload_detached" arg or having the "b64=False" inside the headers.

For decoding, the detached payload content (bytes) has to be provided
with the "detached_payload" arg and "b64=False" has to be found inside
the decoded headers.

Functionnally, when this feature is used, the signature will be computed
over the raw data bytes of the payload, without being base64 encoded and
obviously, the payload will not be provided inside the generated JWS.
So, the generated JWS will look like:
base64url(header)..base64url(signature)

Relevant specifications:
RFC 7515: "JSON Web Signature (JWS)". (Annexe F)
RFC 7797: "JSON Web Signature (JWS) Unencoded Payload Option".

Usage examples:
```
import jwt
payload = b'{'foo':'bar','number':10}'
x_jws = jwt.api_jws.encode(payload, key,  algorithm="HS256", headers=decoded_headers, is_payload_detached=False)
print(x_jws)
# only useful to verify the signature, as, obviously, the payload is already known.
jwt.api_jws.decode(x_jws, key, algorithms=["HS256"], detached_payload=payload)
```

### DIFF
--- a/jwt/api_jws.py
+++ b/jwt/api_jws.py
@@ -80,34 +80,54 @@ class PyJWS:
         algorithm: Optional[str] = "HS256",
         headers: Optional[Dict] = None,
         json_encoder: Optional[Type[json.JSONEncoder]] = None,
+        is_payload_detached: bool = False,
     ) -> str:
         segments = []
 
         if algorithm is None:
             algorithm = "none"
 
-        # Prefer headers["alg"] if present to algorithm parameter.
-        if headers and "alg" in headers and headers["alg"]:
-            algorithm = headers["alg"]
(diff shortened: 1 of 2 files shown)


auvipy:
thanks. restarted the build

auvipy (approved):

fviard:
Hi @jpadilla, is it good for you to merge?
Thanks

auvipy (approved):

PR #746, opened 2022-04-07
Don't raise error when algorithm is not support

oddbeck:
Instead of raising an error for not supported algorithm we log the warning, and continue with other algorithms/keys

### DIFF
--- a/jwt/api_jwk.py
+++ b/jwt/api_jwk.py
@@ -48,6 +48,7 @@ class PyJWK:
 
         if not self.Algorithm:
             import logging
+
             self.key = None
             logger = logging.getLogger(__name__)
             logger.warning("Do not support algorithm %s", algorithm)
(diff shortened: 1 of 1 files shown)


auvipy (changes requested):
check the build failures please

oddbeck:
> 

I'm off for the easter holydays without a computer, will look into it in a week.

jwt/api_jwk.py
```diff
@@ -47,9 +47,13 @@ def __init__(self, jwk_data, algorithm=None):
...
         if not self.Algorithm:
-            raise PyJWKError("Unable to find a algorithm for key: %s" % self._jwk_data)
+            import logging
 
-        self.key = self.Algorithm.from_jwk(self._jwk_data)
+            self.key = None
+            logger = logging.getLogger(__name__)
+            logger.warning("Do not support algorithm %s", algorithm)
+        else:
+            self.key = self.Algorithm.from_jwk(self._jwk_data)
```
akx, line 56: I think this might have security implications, as users of this class likely expect the algorithm in the JWK data to be valid.

akx (changes requested):

DaGuich:
Should be fixed with https://github.com/jpadilla/pyjwt/pull/762 
Sorry, I've seen the PR too late

PR #753, opened 2022-04-28
adding support for compressed payloads

danieltmiles:
A bunch of US states are using Smart Health Cards (https://smarthealth.cards/en/) to issue digital proof of COVID-19 vaccination. The QR code on the certificate decodes to a JWT token, but it has a compressed payload and upstream PyJWT didn't handle it. So I offer this contribution.

This patch introduces a dependency on `zlib` but as an old, salty python developer who hasn't lifted myself out of the requirements.txt era, I haven't figured out how this project is managing dependencies. Can you help me understand what needs to be done?

### DIFF
--- a/tests/test_api_jwt.py
+++ b/tests/test_api_jwt.py
@@ -86,11 +86,13 @@ class TestJWT:
         }
 
     def test_decodes_complete_valid_jwt_with_invalid_compressed_payload(self, jwt):
+        # payload made with the pako (https://nodeca.github.io/pako/) library in Javascript:
+        # Buffer.from(pako.deflateRaw('{"this is": "not valid" json')).toString('base64')
         example_secret = "secret"
         example_jwt = (
             b"eyJhbGciOiAiSFMyNTYiLCAidHlwIjogIkpXVCJ9"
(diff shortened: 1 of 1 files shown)


auvipy (approved):

jpadilla:
Is this something we can instead provide some kind of hook for? This seems kinda specific use case that I’d think would better live outside this lib

jpadilla:
@auvipy let's revert this please. I'd like to think it through a bit more. Sorry for the last minute comments.

auvipy:
I am going to revert this for now, but supporting compressed payload may be a future endeaviour

akx:
It's strange that this was merged when https://github.com/jpadilla/pyjwt/pull/666 is pending to do the same thing.

jpadilla:
@akx we just reverted this in #761 while we figure a better way to allow third parties to implement things like this.

russoale:
@jpadilla Gan you give some update if there is already a way for some sort of callback that can be passed to do this? 
Would be open to help

akx:
@russoale Maybe https://github.com/jpadilla/pyjwt/pull/829 might get merged. :)

PR #762, opened 2022-05-25
Do not fail when an unusable key occurs

DaGuich:
There may be more than keys in the store which may be (still usable).
Therefore we do not want to fail on any key that is not usable by pyjwt,
but rather skip when appending to the KeySet


jpadilla:
>There may be more than keys in the store which may be (still usable).
what's the case for that?

DaGuich:
Keycloak has keys with `"alg": "RSA-OAEP"` which causes loading the keys to fail. But there is also the other key with `"alg": "RSA256"` which are actually use to sign the JWT.

jpadilla:
Great, thanks. Can you please add some simple tests for this?

DaGuich:
I don't know if you want to have further test cases... but I've added some

jpadilla (approved):
looks great, also mind updating CHANGELOG?

jwt/api_jwk.py
```diff
@@ -84,7 +84,14 @@ def __init__(self, keys):
             raise PyJWKSetError("The JWK Set did not contain any keys")
 
         for key in keys:
-            self.keys.append(PyJWK(key))
+            try:
+                self.keys.append(PyJWK(key))
+            except PyJWKError:
+                # skip unusable keys
```
kamil-muszynski-90poe, line 90: Hello, 

I realize this is a bit late since this is merged already, but it produces a behavior like this:

```
  File "/usr/local/lib/python3.8/dist-packages/jwt/jwks_client.py", line 96, in get_signing_key_from_jwt
    return self.get_signing_key(header.get("kid"))
  File "/usr/local/lib/python3.8/dist-packages/jwt/jwks_client.py", line 78, in get_signing_key
    signing_keys = self.get_signing_keys()
  File "/usr/local/lib/python3.8/dist-packages/jwt/jwks_client.py", line 65, in get_signing_keys
    jwk_set = self.get_jwk_set(refresh)
  File "/usr/local/lib/python3.8/dist-packages/jwt/jwks_client.py", line 62, in get_jwk_set
    return PyJWKSet.from_dict(data)
  File "/usr/local/lib/python3.8/dist-packages/jwt/api_jwk.py", line 102, in from_dict
    return PyJWKSet(keys)
  File "/usr/local/lib/python3.8/dist-packages/jwt/api_jwk.py", line 97, in __init__
    raise PyJWKSetError("The JWK Set did not contain any usable keys")
jwt.exceptions.PyJWKSetError: The JWK Set did not contain any usable keys
```
when `PyJWKError` is raised from line 50. This is actually caused by installing pyjwk only, with no crypto alrogithms available, and can be fixed with pip install pyjwk[crypto], as suggested in the docs - but it took me a bit to figure this out.

Maybe it could be possible to at least log a warning that a key was skipped with an error message details?

PR #764, opened 2022-05-26
drop python 3.6 from GA CI

auvipy:


jpadilla (approved):

jpadilla (changes requested):
do we want to update the tox config as well?

PR #781, opened 2022-07-11
Add cacheing functionality for JWK set

wuhaoyujerry:
## Problem
The current PyJWT library only support caching of single signing key. Ref to https://github.com/jpadilla/pyjwt/pull/611

Even though it avoids the network call when the services periodically calling `get_signing_key` with same `kid`, the current implementation doesn't have any TTL, so the cache is stored forever. Also, it still makes network call when different `kid` is provided.

## Solution & Implantation
- Add `JWKSetCache` object that stores jwk set with current timestamp
    - When user try to get signing key from jwk set, the service will try to check if the cache exist and not expired. If so, it will avoid the network call and return the cached jwk set.
    - If the network call throws any error, clear the cache.
    - If the service can't find matching `kid` from the jwk set, it will make a new network call to get a new jwk set and check again in new set. Raise error if still cannot find the signing key with matching kid. 
    - Add option to enable/disable jwk set cache and lifespan. Default lifespan is 5 minutes. 
- Add unit tests to cover the added code
- Disable LRU cache for `get_signing_key` by default as suggested from this comment https://github.com/jpadilla/pyjwt/pull/611#issuecomment-815807580


auvipy (approved):

wuhaoyujerry:
Hi @auvipy, this PR has been approved for a while. Any idea when it can be merged? Let me know if you have any questions.

wuhaoyujerry:
Thx for merging the PR @auvipy, do you have an estimation time on when the new release will be published? My team want to use some features in the master branch and kinda been blocked right now.

jpadilla:
Before releasing this I think we'd need an updated changelog and I think this probably deserves at least _some_ docs.

wuhaoyujerry:
I can add some docs about this functionality

dwyfrequency:
Is there a timeline for when this feature will be released to production?

mactanner:
> I can add some docs about this functionality

Where can the docs be found?
Thanks for this feature.

PR #782, opened 2022-07-21
Update audience typing

JulianMaurin:
Hello :wave: 

Using the library, I just noticed that the typing of `audience` was not that clear.
There is a difference between:
- [annotations](https://github.com/jpadilla/pyjwt/blob/master/jwt/api_jwt.py#L153)
- [type checking](https://github.com/jpadilla/pyjwt/blob/master/jwt/api_jwt.py#L183)
- [documentation](https://github.com/jpadilla/pyjwt/blob/master/docs/api.rst#L65)

This PR aims to bring everything in line.

I removed `bytes` from the type checking, but according to the test it was not handled (raising error [here](https://github.com/jpadilla/pyjwt/blob/master/jwt/api_jwt.py#L260)).

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -62,7 +62,7 @@ API Reference
             if ``verify_exp``, ``verify_iat``, and ``verify_nbf`` respectively
             is set to ``True``).
 
-    :param Iterable audience: optional, the value for ``verify_aud`` check
+    :param Union[str, Iterable] audience: optional, the value for ``verify_aud`` check
     :param str issuer: optional, the value for ``verify_iss`` check
     :param float leeway: a time margin in seconds for the expiration check
     :rtype: dict
--- a/jwt/api_jwt.py
+++ b/jwt/api_jwt.py
@@ -76,7 +76,7 @@ class PyJWT:
         detached_payload: Optional[bytes] = None,
         # passthrough arguments to _validate_claims
         # consider putting in options
-        audience: Optional[str] = None,
+        audience: Optional[Union[str, Iterable]] = None,
         issuer: Optional[str] = None,
         leeway: Union[int, float, timedelta] = 0,
         # kwargs
@@ -150,7 +150,7 @@ class PyJWT:
         detached_payload: Optional[bytes] = None,
         # passthrough arguments to _validate_claims
         # consider putting in options
-        audience: Optional[str] = None,
+        audience: Optional[Union[str, Iterable]] = None,
         issuer: Optional[str] = None,
         leeway: Union[int, float, timedelta] = 0,
         # kwargs
@@ -180,8 +180,8 @@ class PyJWT:
         if isinstance(leeway, timedelta):
             leeway = leeway.total_seconds()
 
-        if not isinstance(audience, (bytes, str, type(None), Iterable)):
-            raise TypeError("audience must be a string, iterable, or None")
+        if audience and not isinstance(audience, (str, Iterable)):
+            raise TypeError("audience must be a string, iterable or None")
 
         self._validate_required_claims(payload, options)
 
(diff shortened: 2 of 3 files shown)


jwt/api_jwt.py
```diff
@@ -180,8 +180,8 @@ def _validate_claims(self, payload, options, audience=None, issuer=None, leeway=
         if isinstance(leeway, timedelta):
             leeway = leeway.total_seconds()
 
-        if not isinstance(audience, (bytes, str, type(None), Iterable)):
-            raise TypeError("audience must be a string, iterable, or None")
+        if audience and not isinstance(audience, (str, Iterable)):
```
JulianMaurin, line 183: Not much faster but I can already tell that it doesn't slow down.

```
from collections.abc import Iterable

def old(audience):
    return not isinstance(audience, (str, type(None), Iterable))

def new(audience):
    return audience and not isinstance(audience, (str, Iterable))
```

```
%timeit old("a")
617 ns ± 53.2 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
%timeit new("a")
513 ns ± 69 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
```
%timeit old(["a"])
1.89 µs ± 361 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
%timeit new(["a"])
1.1 µs ± 78.1 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

```
%timeit old(None)
606 ns ± 91.1 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
%timeit new(None)
223 ns ± 31.2 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

docs/api.rst
```diff
@@ -62,7 +62,7 @@ API Reference
             if ``verify_exp``, ``verify_iat``, and ``verify_nbf`` respectively
             is set to ``True``).
 
-    :param Iterable audience: optional, the value for ``verify_aud`` check
+    :param Union[str, Iterable] audience: optional, the value for ``verify_aud`` check
```
JulianMaurin, line 65: There is no other usage of `Union` in the doc, tell me if you think there is a better way to do it.
(ps: I did not use `Optional` to keep it consistent with the lines bellow)

jwt/api_jwt.py
```diff
@@ -76,7 +76,7 @@ def decode_complete(
         detached_payload: Optional[bytes] = None,
         # passthrough arguments to _validate_claims
         # consider putting in options
-        audience: Optional[str] = None,
+        audience: Optional[Union[str, Iterable]] = None,
```
JulianMaurin, line 79: We can also do :
```suggestion
        audience: Optional[Union[str, Iterable[str]]] = None,
```

But as a `str` is a `Iterable[str]` itself,  I don't know what would be the best. 
This version looks already better to me, so (IMHO) we can go this way.

jwt/api_jwt.py
```diff
@@ -150,7 +150,7 @@ def decode(
...
```
JulianMaurin, line 153: Hey @auvipy, to be consistent with the suggestion applied L79, we can also align this one :) 

```suggestion
        audience: Optional[Union[str, Iterable[str]]] = None,
```

JulianMaurin: Ah. According to the CI, python 3.7 does not like it. 

Maybe we can use `from __future__ import annotations` ?

```
Python 3.7.13
>>> from collections.abc import Iterable
>>> x: Iterable[str] = ["x"]
TypeError: 'ABCMeta' object is not subscriptable
>>> from __future__ import annotations
>>> x: Iterable[str] = ["x"]
>>>

```

JulianMaurin: Done (commits squashed)  :heavy_check_mark:

auvipy (approved):

PR #794, opened 2022-08-24
Handling 'ImmatureSignatureError' for issued_at time

sriharan16:
Handling 'ImmatureSignatureError' for issued_at time when it is a future time epoch.
```
if iat > (now + leeway):
   raise ImmatureSignatureError("The token is not yet valid (iat)")
```

When the `issued_at` time in the payload is greater than the `current time + leeway` then we can call it out as `ImmatureSignatureError` as we do for `(nbf)`

We have `nbf` in the payload but still, with proper `nbf` someone can call the API with improper `iat` to fool the system.

Example:
```
{
  "nbf": 1661419080   # 25-08-2022 14:48 IST
  "iat": 1661419200,    # 25-08-2022 14:50 IST
  "exp": 1661419500   # 25-08-2022 14:55 IST
}
```

Here the token is valid from 14:48(as per `nbf`) and has an expiry range of 5min from `iat` which makes the token valid. But the `iat` is less than `nbf` which makes the token valid for 7mins instead of 5min. This should not happen as per contract but attackers may do something like this even making `iat` and `exp` with the year 2023 which still makes the token valid.

We can restrict the same way as we do for `nbf`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,7 @@ Fixed
 
 Added
 ~~~~~
+- Adding validation for `issued_at` when `iat > (now + leeway)` as `ImmatureSignatureError` by @sriharan16 in https://github.com/jpadilla/pyjwt/pull/794
 
 `v2.5.0 <https://github.com/jpadilla/pyjwt/compare/2.4.0...2.5.0>`__
 -----------------------------------------------------------------------
--- a/jwt/api_jwt.py
+++ b/jwt/api_jwt.py
@@ -210,10 +210,13 @@ class PyJWT:
                 raise MissingRequiredClaimError(claim)
 
     def _validate_iat(self, payload, now, leeway):
+        iat = payload["iat"]
         try:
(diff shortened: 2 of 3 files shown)


sriharan16:
@jpadilla Kindly help here in validating and reviewing this PR.

auvipy (changes requested):
please rebase

sriharan16:
@auvipy I have rebased as mentioned.
Also not sure why requesting your review removed @jpadilla! Kindly verify the changes and also add @jpadilla again as a reviewer

sriharan16:
Team(@jpadilla, @auvipy ),

Can you please review this !

auvipy:
ci triggerred

auvipy (approved):

auvipy:
there is a bug report https://github.com/jpadilla/pyjwt/issues/814 , can you verify?

sriharan16:
Replied there.

PR #805, opened 2022-09-19
Remove `types-cryptography` from `crypto` extra

### LINKED ISSUE #804: `crypto` extra requires outdated `types-cryptography` type annotation stubs
Type annotation stub packages are usually only needed for static type analyzers like `mypy`. Type annotations of dependencies might not be useful in any way for the user of this library if it's not necessary to use the dependency directly. Thus, type annotations of dependencies should be always optional, and definitely not included in the `crypto` extra. Also, annotations in `types-cryptography` are outdated or incomplete for the latest version of `cryptography`. This dependency was added in #784.

`cryptography` package has included type annotations which are usable by `mypy` and other tools since [version 3.4.4](https://github.com/pyca/cryptography/blob/main/CHANGELOG.rst#344---2021-02-09). [Metadata in the `typeshed` repository](https://github.com/python/typeshed/blob/master/stubs/cryptography/METADATA.toml) also suggests that those annotations are obsolete for `cryptography` versions 3.4.4 and later. For some reason, `mypy` prefers the external stubs over the type information from the source code and gives incorrect output for later versions of `cryptography` when `types-cryptography` is installed. For example, a project could use `cryptography`, and also depend on `pyjwt[crypto]`, in which case `mypy` could give incorrect errors for direct use `cryptography`. Thus, I'd suggest not installing `types-cryptography` unless you're running tests and static type analysis against the oldest supported version of `cryptography`.

## Expected Result

`pip install 'pyjwt[crypto]'` should not install `types-cryptography` and `mypy` should not print any errors for the included code snippet.

## Actual Result
`pip install 'pyjwt[crypto]'` installs `types-cryptography` and prints the following error for the code snippet.
```
error: Module "cryptography.hazmat.primitives.ciphers.aead" has no attribute "AESOCB3"
```

## Reproduction Steps

Install `pyjwt` with `crypto` extra, e.g. `pip install 'pyjwt[crypto]'`, the output should show that `types-cryptography` was installed. Run `mypy` on the following code snippet, which simply imports a cipher that was added in `cryptography` version 36.0:
```python
from cryptography.hazmat.primitives.ciphers.aead import AESOCB3
```

## System Information
```json
{
  "cryptography": {
    "version": "38.0.1"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.10.6"
  },
  "platform": {
    "release": "5.19.7-arch1-1",
    "system": "Linux"
  },
  "pyjwt": {
    "version": "2.5.0"
  }
}
```

lautat:
This is not needed for using the project with `cryptography`. Also, the typing information is incorrect or incomplete for the latest version of `cryptography`.

Fixes #804

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -45,7 +45,6 @@ docs =
     zope.interface
 crypto =
     cryptography>=3.3.1
-    types-cryptography>=3.3.21
 tests =
     pytest>=6.0.0,<7.0.0
     coverage[toml]==5.0.4


jpadilla (approved):

jeroenp:
I'd love to see a new release for this so that I don't have to pull in the new dependency types-cryptography when updating to 2.5.0. Thanks.

PR #809, opened 2022-10-01
Differentiate between two errors

irdkwmnsb:
It's impossible to distinguish the two errors when 'audience' doesn't match.
The first error is programmer's fault – supplying an incorrect audience value.
The second is an expected error – when the audience doesn't match the required value.
The error messages shouldn't be the same.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,8 @@ This project adheres to `Semantic Versioning <https://semver.org/>`__.
 Changed
 ~~~~~~~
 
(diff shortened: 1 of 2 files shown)


irdkwmnsb:
Or does the absence of audience from the JWT token should match the absence of audience in the validate call?
Is it possible to not check the audience?

auvipy (approved):

jpadilla (approved):
- Needs updating changelog
- Would changing this be considered a breaking change?

irdkwmnsb:
> * Needs updating changelog
Done.
> * Would changing this be considered a breaking change?
No, since 
1) The error message should not be relied upon error handling
2) If so, this error message already might happen in two different cases, which are indistinguishable, so no logic really aplies there.

auvipy:
yeah agree

PR #821, opened 2022-11-04
fix: use datetime.datetime.timestamp function to have a milliseconds

### LINKED ISSUE #814: The token is not yet valid (iat).
We found an issue in one of our tests in the recent update of yours. Our package works with `pyjwt==2.5.0` but it breaks with `pyjwt==2.6.0`. We will give more details on the issue once we explore it more and will try to provide a minimal example that reproduces the error. However, may I ask you to check if the latests changes could have broken backawards compatibility? We are particularly suspicious of #794. We think this because the error message we get: `The token is not yet valid (iat)`.

We are sorry we could not provide more details yet.

## System Information

    $ python -m jwt.help

```
{
  "cryptography": {
    "version": "38.0.1"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.10.6"
  },
  "platform": {
    "release": "5.19.16-76051916-generic",
    "system": "Linux"
  },
  "pyjwt": {
    "version": "2.6.0"
  }
}
```

daillouf:
Fixes #814

### DIFF
--- a/jwt/api_jwt.py
+++ b/jwt/api_jwt.py
@@ -187,7 +187,7 @@ class PyJWT:
 
         self._validate_required_claims(payload, options)
 
-        now = timegm(datetime.now(tz=timezone.utc).utctimetuple())
+        now = datetime.now(tz=timezone.utc).timestamp()
 
         if "iat" in payload and options["verify_iat"]:
             self._validate_iat(payload, now, leeway)


auvipy (commented):
current tests are passing, but should we try additional tests?

auvipy (approved):

Issue #811, opened 2022-10-05
Request to add an option to not sort headers


BSVogler:
Since #715 the header values are sorted. There is no way to disable that. By upgrading to 2.5.0 this changes the generated token and makes it incompatible with other software we are using.

PR #832, opened 2022-11-28
Add `sort_headers` parameter to `api_jwt.encode`

evroon:
This allows you to not sort headers, which prevents a breaking change between v2.4.0 and v2.5.0

fixes https://github.com/jpadilla/pyjwt/issues/811

Should I create a unittest for this?

### DIFF
--- a/jwt/api_jws.py
+++ b/jwt/api_jws.py
@@ -134,7 +134,6 @@ class PyJWS:
             # True is the standard value for b64, so no need for it
             del header["b64"]
 
-        # Fix for headers misorder - issue #715
         json_header = json.dumps(
             header, separators=(",", ":"), cls=json_encoder, sort_keys=sort_headers
         ).encode()
--- a/jwt/api_jws.py
+++ b/jwt/api_jws.py
@@ -101,6 +101,7 @@ class PyJWS:
         headers: dict[str, Any] | None = None,
         json_encoder: Type[json.JSONEncoder] | None = None,
         is_payload_detached: bool = False,
+        sort_headers: bool = True,
     ) -> str:
         segments = []
 
@@ -135,7 +136,7 @@ class PyJWS:
 
         # Fix for headers misorder - issue #715
         json_header = json.dumps(
-            header, separators=(",", ":"), cls=json_encoder, sort_keys=True
+            header, separators=(",", ":"), cls=json_encoder, sort_keys=sort_headers
         ).encode()
 
         segments.append(base64url_encode(json_header))
--- a/tests/test_api_jws.py
+++ b/tests/test_api_jws.py
@@ -1,5 +1,4 @@
 import json
-from collections import OrderedDict
 from decimal import Decimal
 
 import pytest
@@ -420,7 +419,7 @@ class TestJWS:
         jws_message = jws.encode(
             payload,
             key="\xc2",
-            headers=OrderedDict([("b", "1"), ("a", "2")]),
+            headers=dict([("b", "1"), ("a", "2")]),
             sort_headers=sort_headers,
         )
         header_json = base64url_decode(jws_message.split(".")[0])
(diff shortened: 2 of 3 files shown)


auvipy (changes requested):
it would be great if you can add unit tests as well

evroon:
@auvipy I have added a unittest, but it only makes sure that encoding and decoding is the same when using `sort_headers=True` and `sort_headers=False`, as well as testing that the encoded payload is not the same between those two options. But I don't see a way to check that the headers are properly sorted before the payload is encoded.

jwt/api_jws.py
```diff
@@ -135,7 +136,7 @@ def encode(
 
         # Fix for headers misorder - issue #715
```
akx, line 137: Maybe get rid of this now-irrelevant comment?

akx:
@evroon 
> Do you have an idea to improve the test?

Pass in an `OrderedDict` where you've ensured the keys are in non-sorted order (e.g. `OrderedDict([("b", "1"), ("a", "2")])`), then decode the base64-encoded header and ensure that `.index("b") < .index("a")` in the JSON?

evroon:
> @evroon
> 
> > Do you have an idea to improve the test?
> 
> Pass in an `OrderedDict` where you've ensured the keys are in non-sorted order (e.g. `OrderedDict([("b", "1"), ("a", "2")])`), then decode the base64-encoded header and ensure that `.index("b") < .index("a")` in the JSON?

@akx That sounds like a good idea, but the decoding uses `json.loads`, which returns a regular dict instead of an `OrderedDict`, so I can't assert the ordering after decoding:

https://github.com/evroon/pyjwt/blob/2ccd0e4110aa539163de8c1a8a72c40a1f7aac2e/jwt/api_jws.py#L261-L264

akx:
@evroon You could split the JWT string "by hand" and decode the header segment, no need to use the library's decoder for the test :)

evroon:
Ah yes of course

tests/test_api_jws.py
```diff
@@ -414,6 +415,26 @@ def test_bytes_secret(self, jws, payload):
...
+        )
+
+        message_unsorted_splitted, *_ = jws_message_unsorted.split(".")
+        header_unsorted = base64url_decode(message_unsorted_splitted)
+
+        message_sorted_splitted, *_ = jws_message_sorted.split(".")
+        header_sorted = base64url_decode(message_sorted_splitted)
+
+        assert header_unsorted == b'{"typ":"JWT","alg":"HS256","b":"1","a":"2"}'
+        assert header_sorted == b'{"a":"2","alg":"HS256","b":"1","typ":"JWT"}'
```
akx, line 436: You could use `parametrize` to simplify this test:
```suggestion
    @pytest.mark.parametrize("sort_headers", (False, True))
    def test_sorting_of_headers(self, jws, payload, sort_headers):
        jws_message = jws.encode(
            payload,
            secret="\xc2",
            headers=OrderedDict([("b", "1"), ("a", "2")]),
            sort_headers=sort_headers,
        )

        header_json = base64url_decode(jws_message.split(".")[0])
        assert sort_headers == (header_json.index('"a"') < header_json.index('"b"'))
```

evroon: That's indeed an improvement, I added it in 34db85d

akx (approved):
Looks good to me!

tests/test_api_jws.py
```diff
@@ -1,4 +1,5 @@
 import json
+from collections import OrderedDict
```
auvipy, line 2: can we use simple dict instead of orderedDict here?

evroon: Thanks, I changed it in cf57f33aac5d1c2076abb14ecffdf3967739c5a6

auvipy:
ping me when CI is green

evroon:
@auvipy the CI is green now

auvipy (approved):

evroon:
Thanks!

PR #858, opened 2023-02-09
Removes kid presence verification

dehy:
Per the [RFC 7517](https://www.rfc-editor.org/rfc/rfc7517#section-4.5), the `kid` key is optional.

This PR removes the check for a non-None value.

### DIFF
--- a/jwt/jwks_client.py
+++ b/jwt/jwks_client.py
@@ -74,7 +74,7 @@ class PyJWKClient:
         signing_keys = [
             jwk_set_key
             for jwk_set_key in jwk_set.keys
-            if jwk_set_key.public_key_use in ["sig", None] and jwk_set_key.key_id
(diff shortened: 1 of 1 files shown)


jwt/jwks_client.py
```diff
@@ -74,7 +74,7 @@ def get_signing_keys(self, refresh: bool = False) -> List[PyJWK]:
         signing_keys = [
             jwk_set_key
             for jwk_set_key in jwk_set.keys
-            if jwk_set_key.public_key_use in ["sig", None] and jwk_set_key.key_id
+            if jwk_set_key.public_key_use in ["sig", None]
```
auvipy, line 77: is it possible to test this change?

auvipy (changes requested):

Issue #856, opened 2023-02-08
Type Annotation: key should be of type "str | bytes", not "str".


tholenst:
In various APIs, most notably "api_jwt.encode" and "api_jwt.decode", key is passed in and annotated to be of type "str" ([example](https://github.com/jpadilla/pyjwt/blob/master/jwt/api_jwt.py#L44)).

However, this does not allow all keys to be used. 

For example, using a token generated with https://token.dev/ the following works with a key of types bytes, but not str:

```
import jwt
import base64
token = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.e30.3WyMoLZen_KHpOc7iNDmcYixmUPexwuJpZsVeEjnAA8"
key = base64.b64decode('gICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgIA=')
jwt.decode(token, key, 'HS256')
```

This works as expected, but there is no string which can be used here as the key.

tdg5:
The [docs also suggest in several places](https://pyjwt.readthedocs.io/en/latest/usage.html#encoding-decoding-tokens-with-rs256-rsa) that in some circumstances `key` should be a `cryptography.hazmat.primitives.asymmetric.rsa.RSAPrivateKey`, so seems like this type annotation could use work in that regard as well.

Issue #848, opened 2023-01-08
mypy. Key can be cryptography.hazmat.backends.openssl.rsa._RSAPublicKey not only string


ievgennaida:
Summary.
jwt.decode can accept cryptography.hazmat.backends.openssl.rsa._RSAPublicKey or private key or string.

## Expected Result
Allow to pass not only string.

## Actual Result

Warning from mypy while only string is expected but actually public key is expected.

## Reproduction Steps
jwk_set = PyJWKSet.from_json(data)
jwt.decode("", key= jwk_set.keys[0].key);

## System Information

    $ python -m jwt.help
```
{
  "cryptography": {
    "version": "39.0.0"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.8.10"
  },
  "platform": {
    "release": "10",
    "system": "Windows"
  },
  "pyjwt": {
    "version": "2.6.0"
  }
}
```

Mark-Hetherington:
This tripped us up as well. MyPy suggested we pass a string, but the acceptable type depends on the algorithm, as algorith.prepare_key is called.

This appears to have been discussed in #605, #602 and the type annotation might have been changed to Any. There is more related work happening in #843 and #845

auvipy:
is it fixed now?

Viicos:
> is it fixed now?

I'll probably try to find some time to work on it, but as I said here https://github.com/jpadilla/pyjwt/issues/848#issuecomment-1405133862, it might be tricky

PR #873, opened 2023-04-01
Add complete types to take all allowed keys into account

Viicos:
Fixes https://github.com/jpadilla/pyjwt/issues/602, https://github.com/jpadilla/pyjwt/issues/848, https://github.com/jpadilla/pyjwt/issues/856, https://github.com/jpadilla/pyjwt/issues/864#issuecomment-1458420475

I think this is the best we can get, without having to make use of `TypeVar`/`Generic` for algorithms. Had to tweak some parts to fix `mypy` errors.

One possible thing (as stated in https://github.com/jpadilla/pyjwt/pull/843) could be to add overloads to `encode`. I'll let you decide on this one, as it can hurt code readability.
Regarding the `decode` method, it can't be done as a list of allowed algorithms is passed.

### DIFF
--- a/tests/test_algorithms.py
+++ b/tests/test_algorithms.py
@@ -37,7 +37,7 @@ class TestAlgorithms:
         algo = HMACAlgorithm(HMACAlgorithm.SHA256)
 
         with pytest.raises(TypeError) as context:
-            algo.prepare_key(object())  # type: ignore[type-var]
+            algo.prepare_key(object())  # type: ignore[arg-type]
 
         exception = context.value
         assert str(exception) == "Expected a string value"
@@ -112,7 +112,7 @@ class TestAlgorithms:
         algo = RSAAlgorithm(RSAAlgorithm.SHA256)
 
         with pytest.raises(TypeError):
-            algo.prepare_key(None)
+            algo.prepare_key(None)  # type: ignore[arg-type]
 
     @crypto_required
     def test_rsa_verify_should_return_false_if_signature_invalid(self):
@@ -284,7 +284,7 @@ class TestAlgorithms:
         algo = ECAlgorithm(ECAlgorithm.SHA256)
 
         with pytest.raises(InvalidKeyError):
-            algo.to_jwk({"not": "a valid key"})
+            algo.to_jwk({"not": "a valid key"})  # type: ignore[arg-type]
 
     @crypto_required
     def test_ec_to_jwk_with_valid_curves(self):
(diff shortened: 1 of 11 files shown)


auvipy (changes requested):
can you please check why CI is failing?

Viicos:
> can you please check why CI is failing?

```
  self = <jwt.api_jwt.PyJWT object at 0x7f36e4b70610>
  payload = {'claim': 'insanity', 'exp': 1681192610, 'iss': 'jeff'}
  now = 1681192613.000076, leeway = 3
  
      def _validate_exp(
          self,
          payload: dict[str, Any],
          now: float,
          leeway: float,
      ) -> None:
          try:
              exp = int(payload["exp"])
          except ValueError:
              raise DecodeError("Expiration Time claim (exp) must be an" " integer.")
      
          if exp <= (now - leeway):
  >           raise ExpiredSignatureError("Signature has expired")
  E           jwt.exceptions.ExpiredSignatureError: Signature has expired
  
  jwt/api_jwt.py:304: ExpiredSignatureError
```

Seems like this could fail randomly

tests/test_algorithms.py
```diff
@@ -767,12 +767,6 @@ def test_okp_ed25519_should_reject_non_string_key(self):
         with open(key_path("testkey_ed25519.pub")) as keyfile:
             algo.prepare_key(keyfile.read())
 
-    def test_okp_ed25519_should_accept_unicode_key(self):
```
auvipy, line 770: can you please describe why this test is removed?

auvipy (changes requested):

tests/test_algorithms.py
Viicos: It's a duplicate of

https://github.com/jpadilla/pyjwt/blob/ba726444a6cee75af59feb8ea08294d0ac89bedb/tests/test_algorithms.py#L758-L765

L764-765

In fact I think it's better to move the two valid checks from `test_okp_ed25519_should_reject_non_string_key` to this removed `test_okp_ed25519_should_accept_unicode_key`. Tell me if you want me to do it this way

Viicos:
Most of the remaining `mypy` errors are due to the fact that `alg.prepare_key`/`alg.from_jwk` can return either a public or a private key. Do you want to keep mypy running for tests and have an explicit assignment for the return types of these methods, or just drop `mypy` for tests?

auvipy:
> Do you want to keep mypy running for tests and have an explicit assignment for the return types of these methods

yes

auvipy (approved):

PR #875, opened 2023-04-07
Add a timeout for PyJWKClient requests

daviddavis:
By default, the timeout for urlopen is socket._GLOBAL_DEFAULT_TIMEOUT which is None (meaning that the request never times out):

https://docs.python.org/3/library/socket.html#socket.getdefaulttimeout

This change sets the timeout to 30 but also adds a timeout variable users can set.

### DIFF
--- a/jwt/jwks_client.py
+++ b/jwt/jwks_client.py
@@ -19,12 +19,14 @@ class PyJWKClient:
         cache_jwk_set: bool = True,
         lifespan: int = 300,
         headers: Optional[Dict[str, Any]] = None,
+        timeout: int = 30,
     ):
         if headers is None:
             headers = {}
         self.uri = uri
         self.jwk_set_cache: Optional[JWKSetCache] = None
         self.headers = headers
+        self.timeout = timeout
 
         if cache_jwk_set:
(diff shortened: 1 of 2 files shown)


daviddavis:
Here's a reproducer. Drop this in a file and run it:

```python
from http.server import BaseHTTPRequestHandler, HTTPServer
import time

hostName = "localhost"
serverPort = 8080

class MyServer(BaseHTTPRequestHandler):
    def do_GET(self):
        time.sleep(9999999)
        self.send_response(204)

if __name__ == "__main__":
    webServer = HTTPServer((hostName, serverPort), MyServer)
    print("Server started http://%s:%s" % (hostName, serverPort))

    try:
        webServer.serve_forever()
    except KeyboardInterrupt:
        pass

    webServer.server_close()
```

Then simply use [this example](https://pyjwt.readthedocs.io/en/latest/usage.html#retrieve-rsa-signing-keys-from-a-jwks-endpoint) and set the url to "http://localhost:8080".

auvipy:
would you mind adding the regression test as well?

auvipy (approved):

PR #883, opened 2023-04-24
Inclusive IAT timestamp check.

stephan-nordnes-eriksen:
~~Currently, the iat-check will fail if the IAT timestamp is equal to "now". I believe it should use inclusive test to not fail if it's being validated immediately after being created.~~

Someone more "in-the-know" should validate that the tests and use-case is correct


stephan-nordnes-eriksen:
Sorry, this PR is broken. Closing

PR #884, opened 2023-04-25
Build  and upload PyPI package

jpadilla:
Shamelessly copied from https://github.com/hynek/structlog/blob/main/.github/workflows/pypi-package.yml
All credit goes to @hynek

### DIFF
--- a/.github/workflows/pypi-package.yml
+++ b/.github/workflows/pypi-package.yml
@@ -1,0 +1,69 @@
+---
+name: Build & maybe upload PyPI package
+
+on:
+  push:
+    branches: [main]
+    tags: ["*"]
+  pull_request:
+    branches: [main]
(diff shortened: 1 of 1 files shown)


jpadilla:
@hynek thanks for chiming in! It seems like someone else published the project to the TestPyPI instance https://test.pypi.org/project/PyJWT/ so I think I'm kinda blocked there. Reached out directly to see if it can be transferred 🤷‍♂️

 Re: versioning. gonna take a further look at those.

jpadilla:
@di random question, but know anyone that could look into this https://github.com/pypi/support/issues/2790 ?

hynek:
@di any chance for looking into https://github.com/pypi/support/issues/2891 while we're having fun? 😇

PR #887, opened 2023-05-10
Export PyJWKClientConnectionError class

daviddavis:


c-h-russell-walker (approved):
This looks good - I actually just today while using this library tried to import this exception in order to use in a try/except while debugging connection issues.

So this would be a welcome addition IMHO. 👍

auvipy (approved):

PR #897, opened 2023-06-23
Convert registered claims to string

catalintrif:
RFC 7519: Optional "iss", "sub", "aud" values are case-sensitive strings
https://www.rfc-editor.org/rfc/rfc7519#section-4.1


auvipy (changes requested):
the CI seems to be broken by this pr! can you please investigate?

catalintrif:
> the CI seems to be broken by this pr! can you please investigate?

The CI error "tox-gh-actions couldn't understand the parallel option. ignoring the given option: 0" is related to tox config.
Could you please check if the CI is passing on master?

auvipy (changes requested):
can you please add additional unit tests to verify the changes?

PR #910, opened 2023-08-17
Add support for Python 3.12

hugovk:
The [Python 3.12 release candidate is out!](https://discuss.python.org/t/python-3-12-0-release-candidate-1-released/31137?u=hugovk) :rocket:

> ## Call to action
> 
> We strongly encourage maintainers of third-party Python projects to prepare their projects for 3.12 compatibilities during this phase, and where necessary publish Python 3.12 wheels on PyPI to be ready for the final release of 3.12.0.

Also:

* Drop support for EOL Python 3.7: https://devguide.python.org/versions/
* Upgrade syntax with [pyupgrade](https://github.com/asottile/pyupgrade) --py38-plus
* Only attempt deploy for upstream, otherwise it fails for forks due to missing credentials

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         platform: ["ubuntu-latest", "windows-latest"]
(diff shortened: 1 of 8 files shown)


auvipy (approved):
thank you Hugo!

PR #913, opened 2023-08-23
Update jwt-api to accept either a string or list of strings for issuer validation

mattpollak:
This diff updates jwt issuer validation to accept either a string (as it always has) or a `List[str]`, and consideres the jwt `iss` claim valid if the single string matches or if `jwt[iss]` is found in the list.  Method signatures should be unchanged, though types were updated.


auvipy (approved):

mattpollak:
If there is anything else I can do to get this merged, please let me know!  And thank you for reviewing.

PR #928, opened 2023-11-28
Add openssh private key support

Ansud:
It is mentioned in docs
```
When using the RSASSA-PKCS1-v1_5 algorithms, the key argument in both jwt.encode() and jwt.decode() ("secret" in the examples) is expected to be either an RSA public or private key in PEM or SSH format. 
```
But openssh private keys are not supported. Thus fix it.


Ansud:
nice...

PR #936, opened 2023-12-22
Option to disable refresh/retry on kid mismatch

### LINKED ISSUE #929: Ability to disable refresh & retry on `kid` mismatch
## Description
If the `get_signing_key()` method fails to match the `kid`, then it tries again after refreshing the JWK set. 

I would like the option to disable this behavior; therefore, the method simply throws an exception without retry behavior if the `kid` match is not found.
https://github.com/jpadilla/pyjwt/blob/50ecae21c3ecf36f7c2a285e4e59dae1d9391e3d/jwt/jwks_client.py#L94-L101

## Proposed Solution
Update following methods to like this.
```python
def get_signing_key(self, kid: str, retry: bool) -> PyJWK: 
    ...

def get_signing_key_from_jwt(self, token: str, retry: bool = True) -> PyJWK:
    ...
```

kpark-hrp:
## Description

This PR allows user to selectively disable refresh of signing keys on `kid` mismatch. Fixes #929

## Changes
Default parameter of `retry: bool = True` has been added to two following methods.
```python3
def get_signing_key(self, kid: str, retry: bool = True) -> PyJWK: 
    ...

def get_signing_key_from_jwt(self, token: str, retry: bool = True) -> PyJWK:
    ...
```

**Note: No breaking change as the default behavior remains the same**

### DIFF
--- a/jwt/jwks_client.py
+++ b/jwt/jwks_client.py
@@ -91,26 +91,26 @@ class PyJWKClient:
 
         return signing_keys
 
-    def get_signing_key(self, kid: str) -> PyJWK:
+    def get_signing_key(self, kid: str, retry: bool = True) -> PyJWK:
         signing_keys = self.get_signing_keys()
         signing_key = self.match_kid(signing_keys, kid)
 
-        if not signing_key:
+        if not signing_key and retry:
             # If no matching signing key from the jwk set, refresh the jwk set and try again.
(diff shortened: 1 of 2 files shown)


kpark-hrp:
@jpadilla @auvipy Hi, can you guys take a look at this small PR?

auvipy:
why you closed it?

PR #963, opened 2024-07-05
Raise exception when required cryptography dependency is missing

tobloef:
We recently ran into this when deploying an app with Docker. Locally everything worked fine but the dependency was missing in the container's environment. We had seen the note about it in the docs, but (wrongly) assumed that pyjwt would complain if it tried to use the missing dependency.

My hope with this PR is that it might prevent other people from being stuck on this in the future.

### DIFF
--- a/jwt/api_jwk.py
+++ b/jwt/api_jwk.py
@@ -5,7 +5,13 @@ import time
 from typing import Any
 
 from .algorithms import get_default_algorithms, has_crypto, requires_cryptography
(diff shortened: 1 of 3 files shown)


jpadilla (commented):
Thanks, can you add a test?

tobloef:
> Thanks, can you add a test?

Done!

jpadilla (approved):
🎇 thanks!

PR #970, opened 2024-08-05
chore: use sequence for typing rather than list

imnotjames:
this loosens up the typing for `issuers` & `algorithms` in the API

there's no reason we need to use `List` for the typing of issuers & other parameters - the `Sequence` type allows us to accept more types such as `tuple` and `set` which still support the behavior we need

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,8 @@ This project adheres to `Semantic Versioning <https://semver.org/>`__.
 Changed
 ~~~~~~~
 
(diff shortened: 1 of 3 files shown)


imnotjames:
the reasoning here is that it's a looser type that still supports everything needed with regards to requirements of the parameters

this doesn't change return types to make them looser as that'd be breaking (though, making the return types / exposes prop types sequences & immutable could be a good breaking change?)

jpadilla (commented):
lgtm, thanks! mind updating the CHANGELOG?

imnotjames:
> lgtm, thanks! mind updating the CHANGELOG?

Sure thing, I think I matched previous formats

jpadilla (approved):

PR #979, opened 2024-08-28
Add JWK support to JWT encode

luhn:
To round out my contributions, this PR adds JWK support to `api_jws.encode` and `api_jwt.encode`.

Most significant change is change of default algorithm to `DEFAULT_ALGORITHM`.  This will resolve to the JWK algorithm if applicable, otherwise `HS256`.

### DIFF
--- a/jwt/api_jws.py
+++ b/jwt/api_jws.py
@@ -26,6 +26,9 @@ if TYPE_CHECKING:
     from .algorithms import AllowedPrivateKeys, AllowedPublicKeys
 
 
+DEFAULT_ALGORITHM = "DEFAULT_ALGORITHM"
+
+
 class PyJWS:
     header_typ = "JWT"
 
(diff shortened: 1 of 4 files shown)


jwt/api_jws.py
```diff
@@ -105,8 +108,8 @@ def get_algorithm_by_name(self, alg_name: str) -> Algorithm:
     def encode(
         self,
         payload: bytes,
-        key: AllowedPrivateKeys | str | bytes,
-        algorithm: str | None = "HS256",
+        key: AllowedPrivateKeys | PyJWK | str | bytes,
+        algorithm: str | None = DEFAULT_ALGORITHM,
```
jpadilla, line 112: The `DEFAULT_ALGORITHM` is a bit odd. I wonder if we're better off setting this to None?

jpadilla (changes requested):

jwt/api_jws.py
luhn: I agree it's odd.  However, `None` means "no signature algorithm" outside the context of a PyJWK key, so it's incredibly dangerous to encourage that:  One misstep with the `key` value and all your JWTs are unsigned.

(I wanted to use a sentinel for DEFAULT_ALGORITHM, a la https://peps.python.org/pep-0661/, rather than a string literal but I couldn't find a way to do that that satisfied me.)

luhn: Well maybe “dangerous” is an overstatement.

jpadilla: maybe worth breaking that behavior. if you want an unsecured JWT you'd have to use `algorithm="none"`

luhn: Sounds good to me.

jpadilla (approved):

PR #985, opened 2024-09-12
Fixes #964: Validate key against allowed types for Algorithm family

### LINKED ISSUE #964: Error `TypeError: ECPublicKey.verify() takes 3 positional arguments but 4 were given` while using `jwt.decode`
Getting this error `TypeError: ECPublicKey.verify() takes 3 positional arguments but 4 were given` while trying to decode a token with RS256 algorithm.

## Expected Result

Decode some token via jwt.decode

## Actual Result

```
... (omitted) ...
File /opt/venv/lib/python3.12/site-packages/jwt/api_jwt.py:210, in PyJWT.decode(self, jwt, key, algorithms, options, verify, detached_payload, audience, issuer, leeway, **kwargs)
    203 if kwargs:
    204     warnings.warn(
    205         "passing additional kwargs to decode() is deprecated "
    206         "and will be removed in pyjwt version 3. "
    207         f"Unsupported kwargs: {tuple(kwargs.keys())}",
    208         RemovedInPyjwt3Warning,
    209     )
--> 210 decoded = self.decode_complete(
    211     jwt,
    212     key,
    213     algorithms,
    214     options,
    215     verify=verify,
    216     detached_payload=detached_payload,
    217     audience=audience,
    218     issuer=issuer,
    219     leeway=leeway,
    220 )
    221 return decoded["payload"]

File /opt/venv/lib/python3.12/site-packages/jwt/api_jwt.py:151, in PyJWT.decode_complete(self, jwt, key, algorithms, options, verify, detached_payload, audience, issuer, leeway, **kwargs)
    146 if options["verify_signature"] and not algorithms:
    147     raise DecodeError(
    148         'It is required that you pass in a value for the "algorithms" argument when calling decode().'
    149     )
--> 151 decoded = api_jws.decode_complete(
    152     jwt,
    153     key=key,
    154     algorithms=algorithms,
    155     options=options,
    156     detached_payload=detached_payload,
    157 )
    159 payload = self._decode_payload(decoded)
    161 merged_options = {**self.options, **options}

File /opt/venv/lib/python3.12/site-packages/jwt/api_jws.py:209, in PyJWS.decode_complete(self, jwt, key, algorithms, options, detached_payload, **kwargs)
    206     signing_input = b".".join([signing_input.rsplit(b".", 1)[0], payload])
    208 if verify_signature:
--> 209     self._verify_signature(signing_input, header, signature, key, algorithms)
    211 return {
    212     "payload": payload,
    213     "header": header,
    214     "signature": signature,
    215 }

File /opt/venv/lib/python3.12/site-packages/jwt/api_jws.py:309, in PyJWS._verify_signature(self, signing_input, header, signature, key, algorithms)
    306     raise InvalidAlgorithmError("Algorithm not supported") from e
    307 prepared_key = alg_obj.prepare_key(key)
--> 309 if not alg_obj.verify(signing_input, prepared_key, signature):
    310     raise InvalidSignatureError("Signature verification failed")

File /opt/venv/lib/python3.12/site-packages/jwt/algorithms.py:483, in RSAAlgorithm.verify(self, msg, key, sig)
    481 def verify(self, msg: bytes, key: RSAPublicKey, sig: bytes) -> bool:
    482     try:
--> 483         key.verify(sig, msg, padding.PKCS1v15(), self.hash_alg())
    484         return True
    485     except InvalidSignature:

TypeError: ECPublicKey.verify() takes 3 positional arguments but 4 were given
```

## Reproduction Steps

```python
import jwt

token = "..."
key = "..."
jwt.decode(
  jwt=token,
  key=key
  algorithms=["RS256",]
)
```

## System Information

    $ python -m jwt.help

```
{
  "cryptography": {
    "version": "42.0.5"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.12.4"
  },
  "platform": {
    "release": "5.15.0-113-generic",
    "system": "Linux"
  },
  "pyjwt": {
    "version": "2.8.0"
  }
}
```

pachewise:
- [ ] add tests, make sure current test suite passes
- [ ] remove TODOs

Following discussion in #964, this PR adds checks to make sure the key loaded by any of the `has_crypto = True` Algorithm children is of the correct type for that Algorithm family. I still need to figure out how to check whether it's the right "flavor" of the algo family (for example, that the SHA256 algo is encoding/decoding using a key generated via SHA256).

### DIFF
--- a/jwt/algorithms.py
+++ b/jwt/algorithms.py
@@ -770,8 +770,9 @@ if has_crypto:
             pass
 
         def prepare_key(self, key: AllowedOKPKeys | str | bytes) -> AllowedOKPKeys:
-            if isinstance(key, self._crypto_key_types):
-                return key
+            if not isinstance(key, (str, bytes)):
+                self.check_crypto_key_type(key)
+                return cast("AllowedOKPKeys", key)
 
             key_str = key.decode("utf-8") if isinstance(key, bytes) else key
             key_bytes = key.encode("utf-8") if isinstance(key, str) else key
--- a/jwt/algorithms.py
+++ b/jwt/algorithms.py
@@ -81,6 +81,7 @@ except ModuleNotFoundError:
 
 if TYPE_CHECKING:
     from typing import TypeAlias
+    from cryptography.hazmat.primitives.asymmetric.types import PublicKeyTypes, PrivateKeyTypes
     # Type aliases for convenience in algorithms method signatures
     AllowedRSAKeys: TypeAlias = RSAPrivateKey | RSAPublicKey
     AllowedECKeys: TypeAlias = EllipticCurvePrivateKey | EllipticCurvePublicKey
@@ -152,7 +153,7 @@ class Algorithm(ABC):
     """
 
     # pyjwt-964: Validate to ensure the key passed in was decoded to the correct cryptography key family
-    _crypto_key_types: tuple[AllowedKeys, ...] = None
+    _crypto_key_types: tuple[type[AllowedKeys], ...] | None = None
 
     def compute_hash_digest(self, bytestr: bytes) -> bytes:
         """
@@ -176,7 +177,7 @@ class Algorithm(ABC):
         else:
             return bytes(hash_alg(bytestr).digest())
 
-    def check_crypto_key_type(self, key: Any):
+    def check_crypto_key_type(self, key: PublicKeyTypes | PrivateKeyTypes):
         """Check that the key belongs to the right cryptographic family.
 
         Note that this method only works when cryptography is installed.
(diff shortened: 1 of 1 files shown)


pachewise:
@auvipy I think this is ready for review, though I'm working on the tests today. Been a while since I opened this PR, so some decisions may not be as good anymore; generally, I tried to keep the same "logic" of the code (keeping multiple returns, etc) while just inserting the `self.check_crypto_key_type()` before returning.

jwt/algorithms.py
```diff
@@ -163,6 +199,30 @@ def compute_hash_digest(self, bytestr: bytes) -> bytes:
         else:
             return bytes(hash_alg(bytestr).digest())
 
+    def check_crypto_key_type(self, key: PublicKeyTypes | PrivateKeyTypes):
```
pachewise, line 202: I also considered putting this in a sub-ABC "CryptoAlgorithm", but that seemed like more changes for not as much value. Let me know if you'd rather I do that instead, or if you would prefer an alternate approach.

jwt/algorithms.py
```diff
@@ -163,6 +199,31 @@ def compute_hash_digest(self, bytestr: bytes) -> bytes:
...
+        Raises:
+            ValueError: if `cryptography` is not installed, or this method is called by a non-cryptography algorithm
+            InvalidKeyError: if the key doesn't match the expected key classes
+        """
+        if not has_crypto or self._crypto_key_types is None:
+            raise ValueError(
+                "This method requires the cryptography library, and should only be used by cryptography-based algorithms."
+            )
+
+        # TODO check for algo_type? (e.g., SHA256 vs SHA384)
```
pachewise, line 218: this is more of a question (should we check algo type?), since the current checks don't do that

auvipy: will that add more value?

auvipy: I guess we should try

pachewise: hmm, @auvipy I gave this a try, but I couldn't find an easy way to determine hash type to compare to in this case. I also don't remember what my initial idea was for this (maybe I thought it'd be easier). I vote we skip this for now.

auvipy (approved):

PR #986, opened 2024-09-12
fix doctest for docs/usage.rst

pachewise:
Breaking CI checks for #985, and probably `master` branch CI checks.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -310,7 +310,7 @@ Retrieve RSA signing keys from a JWKS endpoint
     ...     signing_key,
     ...     audience="https://expenses-api",
     ...     options={"verify_exp": False},
-    ...     algorithms=["RS256"])
+    ...     algorithms=["RS256"],
     ... )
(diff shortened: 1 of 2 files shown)


pachewise:
@jpadilla it looks like tox-docs run is failing because cryptograph isn't installed in that build. Should we be installing cryptograph in order to run the docs tests? need some guidance on how you'd like me to fix this

auvipy (approved):

auvipy:
> @jpadilla it looks like tox-docs run is failing because cryptography isn't installed in that build. Should we be installing cryptography in order to run the docs tests? need some guidance on how you'd like me to fix this

you could try and see

pachewise:
@auvipy I added the change to install `cryptography` on the `docs` tox-env. I am a little worried that this is deviating from the `readthedocs.yaml`, but I don't see a way to fix these tests without splitting up the test runs. I am also curious why the _sphinx_ doc-test is passing given some of the examples require `cryptography` to run.

jpadilla (approved):

PR #987, opened 2024-09-12
fix test_utils.py not to xfail

pachewise:
See https://docs.pytest.org/en/stable/example/parametrize.html#parametrizing-conditional-raising. We should be "passing" the test if the correct action is to raise a ValueError.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,3 +1,4 @@
+from contextlib import nullcontext
 import pytest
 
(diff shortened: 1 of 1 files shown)


auvipy (changes requested):
thanks for the patch. can you please check the test failures?

pachewise:
@auvipy the issue is the one we fixed in #986; once we merge that, the tests should pass here. 
image

jpadilla (approved):

auvipy (approved):

PR #994, opened 2024-10-01
Correct jwt.decode audience param doc expression

peter279k:
As title, correcting the `audience` parameter type expression in the `jwt.decode` method.

Before:

image

After:

image


jpadilla (approved):
thank you!

PR #1008, opened 2024-10-22
Add missing use of "typing.Literal"

kkirsche:
Add missing use of `typing.Literal` in jwt.utils.py

This feature was added in Python 3.8 per https://docs.python.org/3/library/typing.html#typing.Literal


kkirsche:
Oops, I misread this, this is a postponed annotation

PR #1045, opened 2025-03-01
Fixes #660: Improve typing/logic for `options` in decode, decode_complete; Improve docs

### LINKED ISSUE #660: Published types aren't complete enough for Pyright/Pylance
It seems a bit unfair to file this as a "bug," when really what's going on is that the Python community is trying to figure out what a "typed" Python library looks like. In this case, what looks like perfectly reasonable code is, in another light, not explicit enough to allow tools to determine types.

(For clarity's sake: Pylance is Microsoft's new language-analysis extension for Python in VS Code. Pylance is closed source, but wraps the open source type engine Pyright.)

## Expected Result

I have a repo at https://github.com/fluggo/pylance-test that demonstrates this issue. Pylance 2021.5.3 [now knows to bypass the stub libraries](https://github.com/microsoft/pylance-release/issues/1197) when the installed library uses `py.typed` to declare itself typed, and so it should be able to see that this call to encode is wrong:

```python
import jwt

# This should be an error because payload expects Dict[str, Any]
jwt.encode(payload='subsandwich')
```

## Actual Result

Unfortunately it doesn't, because according to Pyright, the type is unknown:

image

This is because Pyright has decided to [play it safe with type inferences](https://github.com/microsoft/pyright/issues/1846). The problematic code is at the bottom of api_jwt.py:

```python
_jwt_global_obj = PyJWT()
encode = _jwt_global_obj.encode
decode_complete = _jwt_global_obj.decode_complete
decode = _jwt_global_obj.decode
```

These assignments look cut and dry—PyJWT is just creating a module-level alias for these functions. But it only looks that way to a programmer—putting our language lawyer hats on, these are variables, not constants, and they can be reassigned. Since that's the case, what are the other valid values for these variables? Pyright can't know, and can't assume they will be constrained to the type of their first assignment.

There's also the issue that, if an assignment like this requires any type inference at all, the results could be different from one analysis tool to the next. This is different from the situation in TypeScript, because in TypeScript, there's only the one implementation of type inference, whereas Python doesn't yet have a standard for inferring types.

In Pyright's analysis, these variables need their own type annotations, and they recommend so in their [Typing Guidance for Python Libraries](https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md).

The discussion on what to do here is still ongoing in microsoft/pyright#1846, but as things sit now, this is something library implementors will have to wrestle with. There's also a bit more background in microsoft/pylance-release#1197.

## Reproduction Steps

Grab my repo above and look at it with Pylance under VS Code. Apologies to anyone not using Poetry; I'm not well versed with the Python package managers.

## System Information

    $ python -m jwt.help

```
{
  "cryptography": {
    "version": "3.4.7"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.9.4"
  },
  "platform": {
    "release": "19.6.0",
    "system": "Darwin"
  },
  "pyjwt": {
    "version": "2.1.0"
  }
}
```

pachewise:
Basically trying to remove "Unknown" or "partially unknown" pylance issues. Mostly with `options`, but there were a few simple other fixes (kwargs, segments, etc). This showed some minor bugs and a few code smells (passing too many args to `api_jws.decode_complete`, for example).

- [x] We should include Options / SigOptions / FullOptions in the docs somehow
- [x] `PyJWT()` is used as a singleton. Do we want to keep the `options` param in that case? Wondering because codecov is warning that we don't have a test for `options is not None` case.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,7 +18,7 @@ Added
 ~~~~~
 
 - Docs: Add example of using leeway with nbf by @djw8605 in `#1034 <https://github.com/jpadilla/pyjwt/pull/1034>`__
-- Docs: Refactored docs with ``autodoc``, added ``PyJWS`` and ``jwt.algorithms`` docs by @pachewise in `#1045 <https://github.com/jpadilla/pyjwt/pull/1045>`__
+- Docs: Refactored docs with ``autodoc``; added ``PyJWS`` and ``jwt.algorithms`` docs by @pachewise in `#1045 <https://github.com/jpadilla/pyjwt/pull/1045>`__
 
 `v2.10.1 <https://github.com/jpadilla/pyjwt/compare/2.10.0...2.10.1>`__
 -----------------------------------------------------------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,7 @@ Added
 ~~~~~
 
 - Docs: Add example of using leeway with nbf by @djw8605 in `#1034 <https://github.com/jpadilla/pyjwt/pull/1034>`__
+- Docs: Refactored docs with ``autodoc``, added ``PyJWS`` and ``jwt.algorithms`` docs by @pachewise in `#1045 <https://github.com/jpadilla/pyjwt/pull/1045>`__
 
 `v2.10.1 <https://github.com/jpadilla/pyjwt/compare/2.10.0...2.10.1>`__
 -----------------------------------------------------------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ Fixed
 - Validate key against allowed types for Algorithm family in `#964 <https://github.com/jpadilla/pyjwt/pull/964>`__
 - Add iterator for JWKSet in `#1041 <https://github.com/jpadilla/pyjwt/pull/1041>`__
 - Validate `iss` claim is a string during encoding and decoding by @pachewise in `#1040 <https://github.com/jpadilla/pyjwt/pull/1040>`__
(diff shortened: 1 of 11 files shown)


jwt/api_jwt.py
```diff
@@ -236,9 +261,9 @@ def decode(
     def _validate_claims(
         self,
         payload: dict[str, Any],
-        options: dict[str, Any],
-        audience=None,
-        issuer=None,
+        options: FullOptions,
+        audience: Iterable[str] | str | None = None,
+        issuer: Iterable[str] | str | None = None,
```
pachewise, line 266:
```suggestion
        issuer: Container[str] | str | None = None,
```
see discussion in #1040

auvipy (commented):
ping me when it is ready for review

pachewise:
@auvipy this should be ready to review. Note that I changed the documentation to use docstrings/autodoc so that the docs don't get stale (I saw a few cases where that happened). Also added more docs where needed; let me know if you'd rather me separate that into another PR. 

I still need an answer on whether we want to support the `PyJWT(options)` call. If no opinion, I will leave it untouched for now and add a test for the missing branch (see codecov failure)

CHANGELOG.rst
```diff
@@ -11,7 +11,7 @@ Fixed
 ~~~~~
 - Validate key against allowed types for Algorithm family in `#964 <https://github.com/jpadilla/pyjwt/pull/964>`__
 - Add iterator for JWKSet in `#1041 <https://github.com/jpadilla/pyjwt/pull/1041>`__
-
+- Improve typing/logic for `options` in decode, decode_complete by @pachewise in `#1045 <https://github.com/jpadilla/pyjwt/pull/1045>`__
```
pachewise, line 14: In Added section, include that we added docs for Algorithm, types, etc.

auvipy (changes requested):
can you please fix the merge conflicts?

auvipy (approved):

PR #1046, opened 2025-03-10
Fix TYP header documentation

fobiasmog:
Hey, spent half an hour while debugging why my token for APNS is not the same length like generated from apple :) And finally found, that in my token header "typ": "JWT" is included by default, so probably it will be good to reflect this in documentation :)


auvipy (approved):

auvipy (commented):
any reference to it?

fobiasmog:
@auvipy what reference do you mean? I didn't make any Issue btw

PR #1048, opened 2025-03-17
Fixes #1047: Correct `PyJWKClient.get_signing_key_from_jwt` annotation 

### LINKED ISSUE #1047: Incomplete `PyJWKClient.get_signing_key_from_jwt` type annotation
`PyJWKClient.get_signing_key_from_jwt` method is not correctly type annotated. Even though the underlying code expects `str | bytes`, the `jwt` argument is only annotated with `str`. 

## Expected Result

A `def get_signing_key_from_jwt(self, token: str | bytes) -> PyJWK:` annotation, allowing the library users without having to extraneously decode the jwt.

Ideally, a single `JWT: TypeAlias = str | bytes` type exported for the library users and used in all relevant annotations.

khvn26:
Fixes #1047.

### DIFF
--- a/jwt/jwks_client.py
+++ b/jwt/jwks_client.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import json
 import urllib.request
 from functools import lru_cache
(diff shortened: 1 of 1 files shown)


auvipy (changes requested):
can you please check why the builds are failing?

auvipy (changes requested):
I can see type error for this pr

PR #1050, opened 2025-03-18
Removed redundant args in `_encode_payload` method

KrishnenduDG:
Well, was just going through the basic implementation of the whole jwt flow. Just then I realised that there was an extra and unused argument headers being passed in the _encode_payload method. Removed that from both the function definition and the function call.


auvipy (changes requested):
how can we verify this change?

Issue #1063, opened 2025-05-25
Doc string typo fix for PyJWT._validate_jti() function


kuldeepkhatke:
Just a simple Doc string typo fix required for PyJWT._validate_jti() function

## Expected Result
```
def _validate_jti(self, payload: dict[str, Any]) -> None:
        '''
        Checks whether "jti" if in the payload is valid **or** not
        This is an Optional claim

        :param payload(dict): The payload which needs to be validated
        '''
```

## Actual Result

```
def _validate_jti(self, payload: dict[str, Any]) -> None:
        '''
        Checks whether "jti" if in the payload is valid **ot** not
        This is an Optional claim

        :param payload(dict): The payload which needs to be validated
        '''
```

## Reproduction Steps

Check at pyjwt\jwt\api_jwt.py:439

## System Information

NA

auvipy:
feel free to contribute

PR #1064, opened 2025-05-25
Fixed doc string typo in _validate_jti() function #1063

kuldeepkhatke:
Fixed docstring typo 

## Current Result
```
def _validate_jti(self, payload: dict[str, Any]) -> None:
        '''
        Checks whether "jti" if in the payload is valid **or** not
        This is an Optional claim

        :param payload(dict): The payload which needs to be validated
        '''
```

## Previous Result

```
def _validate_jti(self, payload: dict[str, Any]) -> None:
        '''
        Checks whether "jti" if in the payload is valid **ot** not
        This is an Optional claim

        :param payload(dict): The payload which needs to be validated
        '''
```

### DIFF
--- a/jwt/api_jwt.py
+++ b/jwt/api_jwt.py
@@ -436,7 +436,7 @@ class PyJWT:
 
     def _validate_jti(self, payload: dict[str, Any]) -> None:
         """
-        Checks whether "jti" if in the payload is valid ot not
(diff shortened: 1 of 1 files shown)


kuldeepkhatke:
Hi @jpadilla , 
Please review & merge this quick fix. 
Thanks

PR #1068, opened 2025-06-12
Typing fix: use `float` instead of `int` for `lifespan` and `timeout`

nikitagashkov:
This MR updates the type hints for `lifespan` and `timeout` parameters from `int` to `float` in `JWKSetCache` and `PyJWKClient`.

Timeouts are commonly specified as `float`s to allow for sub-second precision (e.g., `timeout=0.5`). Updating the type hints avoids unnecessary type checker warnings when `float` values are passed.

This change is backwards compatible since in the type system [`float`s allow `int`s to be accepted](https://typing.python.org/en/latest/spec/special-types.html#special-cases-for-float-and-complex) and it is actually considered a [best bractice](https://typing.python.org/en/latest/reference/best_practices.html#types).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ Fixed
 - Add iterator for JWKSet in `#1041 <https://github.com/jpadilla/pyjwt/pull/1041>`__
 - Validate `iss` claim is a string during encoding and decoding by @pachewise in `#1040 <https://github.com/jpadilla/pyjwt/pull/1040>`__
 - Improve typing/logic for `options` in decode, decode_complete by @pachewise in `#1045 <https://github.com/jpadilla/pyjwt/pull/1045>`__
+- Declare float supported type for lifespan and timeout by @nikitagashkov in `#1068 <https://github.com/jpadilla/pyjwt/pull/1068>`__
 
 Added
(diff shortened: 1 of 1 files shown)


jwt/jwks_client.py
```diff
@@ -20,9 +20,9 @@ def __init__(
         cache_keys: bool = False,
         max_cached_keys: int = 16,
         cache_jwk_set: bool = True,
-        lifespan: int = 300,
+        lifespan: float = 300,
```
auvipy, line 23: is 300 a int or float?

auvipy (changes requested):

jwt/jwks_client.py
ods: > Then why using float here?

Maybe to avoid code like this:

```python
KEYCLOAK_CERTS_URL = "..."
KEYCLOAK_TIMEOUT = 0.5
jwks_client = jwt.PyJWKClient(
    KEYCLOAK_CERTS_URL,
    # Error caused by typing bug in `pyjwt`
    timeout=KEYCLOAK_TIMEOUT,  # type: ignore[arg-type]
)
```

auvipy (approved):

PR #1073, opened 2025-06-25
Allow validating JWT against JWKS without kid

### LINKED ISSUE #1072: Support JWTs with no kid in the header
## Summary.

I've hit a problem in the wild with JWTs being signed without a `kid` and the corresponding JWK also not containing a `kid`. This appears to be a valid use case:

[RFC 7517 4.5](https://datatracker.ietf.org/doc/html/rfc7517#section-4.5)

> The "kid" (key ID) parameter is used to match a specific key.  This
   is used, for instance, to choose among a set of keys within a JWK Set
   during key rollover.
> ...
> Use of this member is OPTIONAL.

IE it's only functionally necessary to include one when there is some ambiguity around which key. Eg: during key rollover. At other times, when there is only one key, neither the JWT nor the JWK need to include one.

It appears to be valid to match a token with no `kid` against any signing key on a JWK. There is no specification for what to do in the case of ambiguity that I can see, but it seems reasonable to fail validation in that situation.

## What you expected.

If there is not `kid` in the token header and there is exactly one signing key on the JWK then the following code should still work:

```python
jwks_client = jwt.jwks_client.PyJWKClient(config.jwk_url)
signing_key = jwks_client.get_signing_key_from_jwt(token)
decoded = jwt.decode_complete(token, signing_key)
```

This should:
 - Fetch the JWK
 - Filter a list of signing keys _(including keys with no `kid`)_
 - If the token contains a `kid` it should match on the `kid`
 - If the token does not contain a `kid` it verify the list of signing keys contains just one key and verify against that.

## System Information

    $ python -m jwt.help

```
{
  "cryptography": {
    "version": "45.0.4"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.12.0"
  },
  "platform": {
    "release": "24.5.0",
    "system": "Darwin"
  },
  "pyjwt": {
    "version": "2.10.1"
  }
}

```

This appears to be a related to #857

couling:
Fixes #1072 

`get_signing_key` Now accepts `None` for the `kid` and will match to the only JWKS signing key if there is only one. Otherwise it will not match.

`get_signing_keys` Now returns keys with a `None` `kid` to support this.

`get_signing_key_from_jwt` Now defaults the `kid` to `None` when it looks for it in the token.

`match_kid` now accepts None likewise.

### DIFF
--- a/jwt/jwks_client.py
+++ b/jwt/jwks_client.py
@@ -114,7 +114,7 @@ class PyJWKClient:
     def get_signing_key_from_jwt(self, token: str | bytes) -> PyJWK:
         unverified = decode_token(token, options={"verify_signature": False})
         header = unverified["header"]
-        return self.get_signing_key(header.get("kid", None))
+        return self.get_signing_key(header.get("kid"))
 
     @staticmethod
     def match_kid(signing_keys: List[PyJWK], kid: Optional[str]) -> Optional[PyJWK]:


couling:
I've converted this back to a draft. For now. I've just spotted something stranger in the example I'm seeing in the wild and I want to be sure that both the issue #1072 and this PR are capturing the correct behaviour. Both for RFC standard and real working example.

PR #1082, opened 2025-08-08
feat: strict encode, lenient decode for JWT subject claim

gshmu:
Apply "strict with self, lenient with others" philosophy to JWT subject validation:

- **Encode (strict)**: Validate `sub` must be string, raise `TypeError` if not
- **Decode (lenient)**: Remove string type validation for `sub` claim in `_validate_sub`
- **Tests**: Update test to expect `TypeError` during encoding instead of decode error

Philosophy

This change embodies the principle of "严于律己，宽以待人" (be strict with yourself, lenient with others):

- **Strict encoding**: Ensures PyJWT generates standards-compliant tokens
- **Lenient decoding**: Allows PyJWT to consume tokens from other systems that may use non-string subject identifiers

### DIFF
--- a/jwt/api_jwt.py
+++ b/jwt/api_jwt.py
@@ -127,6 +127,10 @@ class PyJWT:
         if "iss" in payload and not isinstance(payload["iss"], str):
             raise TypeError("Issuer (iss) must be a string.")
 
+        # Validate sub must be string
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
please fix the merge conflicts

PR #1109, opened 2025-10-22
Add pyupgrade as a pre-commit hook

kurtmckee:
This introduces pyupgrade as a pre-commit hook, then runs `pre-commit run -a` and commits the changes introduced.

The changes involve one of these two categories of changes:

* `typing.X` is a deprecated alias for `collections.abc.X`
* Type annotations no longer require `Dict`/`List`/etc, and can be replaced with `dict`/`list`/etc.

### DIFF
--- a/jwt/api_jwk.py
+++ b/jwt/api_jwk.py
@@ -2,7 +2,8 @@ from __future__ import annotations
 
 import json
 import time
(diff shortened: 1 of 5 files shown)


kurtmckee:
_Rebased on `master` and resolved merge conflicts._

PR #1127, opened 2025-12-25
Fix #1126: Fix microseconds lose on jwt.encode for known time claims

### LINKED ISSUE #1126: jwt.encode loses microseconds precision when using datetime for known time claims
## Summary

When using `jwt.encode` function, known time claims such as `exp`, `iat` and `nbf` lose microseconds precision if their value is a datetime.

This happens because the underlying structure used by `utctimetuple` and `timegm` functions does not store time units below full seconds (see `tm` struct at https://pubs.opengroup.org/onlinepubs/9799919799/).

Culprit is line at https://github.com/jpadilla/pyjwt/blob/dd448344c3eb7bc54b7b395e0f6b00bb811be24a/jwt/api_jwt.py#L127-L130

## Example

Code

```python
import datetime as dt
import jwt

dt_iat = dt.datetime(2025, 1, 1, 0, 0, 0, 123456, dt.UTC)

result = jwt.decode(jwt.encode({'iat': dt_iat}, 'secret'), 'secret', algorithms = ['HS256'])
decoded_dt_iat = dt.datetime.fromtimestamp(result['iat'], tz=dt.UTC)

print(f'Original: {dt_iat}')
print(f'Decoded:  {decoded_dt_iat}')

print(f'Original timestamp: {dt_iat.timestamp()}')
print(f'Decoded  timestamp: {decoded_dt_iat.timestamp()}')
```

returns
```
Original: 2025-01-01 00:00:00.123456+00:00
Decoded:  2025-01-01 00:00:00+00:00
Original timestamp: 1735689600.123456
Decoded  timestamp: 1735689600.0
```

currently at master (dd448344c3eb7bc54b7b395e0f6b00bb811be24a).

It should return
```
Original: 2025-01-01 00:00:00.123456+00:00
Decoded:  2025-01-01 00:00:00.123456+00:00
Original timestamp: 1735689600.123456
Decoded  timestamp: 1735689600.123456
```

## System Information

    $ python -m jwt.help

```
{
  "cryptography": {
    "version": "46.0.3"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.13.11"
  },
  "platform": {
    "release": "6.12.63-1-lts",
    "system": "Linux"
  },
  "pyjwt": {
    "version": "2.10.1"
  }
}
```

---

I will submit a PR shortly to tackle this.

rodrigovillarbello:
## Summary

Fixes #1126 by using the native `timestamp` package function to convert the datetime directly to a float number, avoiding the lose of microseconds precision.

A very similar change was merged at #821.

Converting directly to float should be no problem since you can already bypass this limitation by converting the datetimes to float yourselt. Code
```python
import datetime as dt
import jwt

ts_iat = dt.datetime(2025, 1, 1, 0, 0, 0, 123456, dt.UTC).timestamp()

result = jwt.decode(jwt.encode({'iat': ts_iat}, 'secret'), 'secret', algorithms = ['HS256'])
decoded_ts_iat = result['iat']

print(f'Original timestamp: {ts_iat}')
print(f'Decoded  timestamp: {decoded_ts_iat}')
```
produces
```
Original timestamp: 1735689600.123456
Decoded  timestamp: 1735689600.123456
```

### DIFF
--- a/jwt/api_jwt.py
+++ b/jwt/api_jwt.py
@@ -125,9 +125,9 @@ class PyJWT:
         # Payload
         payload = payload.copy()
         for time_claim in ["exp", "iat", "nbf"]:
-            # Convert datetime to a intDate value in known time-format claims
+            # Convert datetime to a float value in known time-format claims
             if isinstance(payload.get(time_claim), datetime):
-                payload[time_claim] = timegm(payload[time_claim].utctimetuple())
+                payload[time_claim] = payload[time_claim].timestamp()
 
         # Issue #1039, iss being set to non-string
         if "iss" in payload and not isinstance(payload["iss"], str):
--- a/jwt/api_jwt.py
+++ b/jwt/api_jwt.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import json
 import os
 import warnings
-from calendar import timegm
 from collections.abc import Container, Iterable, Sequence
 from datetime import datetime, timedelta, timezone
 from typing import TYPE_CHECKING, Any, Union, cast


auvipy (changes requested):
please also add proper test

rodrigovillarbello:
Is this still needed? Seems like https://github.com/jpadilla/pyjwt/commit/74961f165765300dc6b366c6c2b3dc0d11b92bde should work

PR #1130, opened 2026-01-30
Fix invalid URL handling in PyJWKClient.fetch_data()

Kontrolix:
### Problem
When `PyJWKClient` is initialized with an invalid URL string and `fetch_data()` is called, `urllib.request.Request` raises a `ValueError`, which is not caught or handled gracefully.

### Solution
- Isolated URL validation into a separate try-except block to explicitly catch `ValueError` when constructing the request
- Converts the `ValueError` to `PyJWKClientError` with a descriptive message that includes the invalid URL and the original error
- Prevents `None` from being cached on invalid URLs (by returning early)

### Testing
- Added `test_fetch_data_invalid_url()` to verify that invalid URLs raise `PyJWKClientError` with the appropriate error message

### Benefits
- Clearer error handling for invalid URL configurations
- Consistent error types across the API (all errors raised as `PyJWKClientError`)
- Better debugging experience with detailed error messages

### DIFF
--- a/jwt/jwks_client.py
+++ b/jwt/jwks_client.py
@@ -54,6 +54,10 @@ class PyJWKClient:
         jwk_set: Any = None
         try:
             r = urllib.request.Request(url=self.uri, headers=self.headers)
+        except ValueError as e:
+            raise PyJWKClientError(f'Invalid URL "{self.uri}", err: "{e}"') from e
+
+        try:
             with urllib.request.urlopen(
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
please fix the merge conflict

Kontrolix:
The URI scheme check now done in PyJWKClient.__init__ rejects the invalid URLs this PR handled, so the ValueError catch in fetch_data can no longer be reached in normal use.

I guess we can close it.

PR #1159, opened 2026-04-19
Keep the JWK Set cache when a refresh fetch fails

Nadav0077:
PyJWKClient.fetch_data currently stores the parsed JWK Set inside a finally clause. When the request raises (URLError, TimeoutError, or a json.load ValueError on a bad response), the local jwk_set is still None, so the finally branch calls jwk_set_cache.put(None) and clears the cache. Any brief outage on the JWKS endpoint wipes a perfectly
valid cached set, forces the next call back onto the network, and turns a single transient failure into a cascading one while the endpoint is recovering.

The fix is to only write to the cache after the parse succeeds. On failure the previously cached set is left untouched and the next non-refresh call is still served from memory.

While looking into this I noticed the existing test, test_get_jwt_set_failed_request_should_clear_cache, wasn't actually
pinning anything. Its only assertion sits inside the pytest.raises block, after a line that always raises, so control leaves the block before the assert executes. The attribute it checks (jwks_client.jwk_set_cache) is the JWKSetCache wrapper, which is
never None once cache_jwk_set is enabled, so even if it did run it wasn't looking at the right thing. I replaced it with
test_get_jwt_set_failed_request_preserves_cache, which primes the cache, triggers a failed refresh, then checks the cached payload is still there and that a follow-up get_jwk_set() doesn't make a new request.

Full test suite passes locally (344 passed, 4 skipped without cryptography installed).

### DIFF
--- a/jwt/jwks_client.py
+++ b/jwt/jwks_client.py
@@ -97,12 +97,12 @@ class PyJWKClient:
 
         Makes an HTTP request to the configured ``uri`` and returns the
         parsed JSON response. If the JWK Set cache is enabled, the
-        response is stored in the cache.
+        response is stored in the cache on success; a previously cached
+        set is preserved if the fetch fails.
 
         :returns: The parsed JWK Set as a dictionary.
         :raises PyJWKClientConnectionError: If the HTTP request fails.
         """
-        jwk_set: Any = None
         try:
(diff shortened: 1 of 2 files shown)


jpadilla:
Addressed in 2.13.0

PR #1169, opened 2026-05-20
fix(jwks_client): Prevent SSRF by validating URI scheme

chu0119:
## Summary

This PR fixes a Server-Side Request Forgery (SSRF) vulnerability in  where attacker-controlled JWKS URIs could use non-HTTP protocols to access unintended resources.

## Problem

 uses  which supports multiple URL schemes including , , , and . When the JWKS URI is attacker-controlled (e.g., via JWT  header in multi-tenant applications), this can be exploited to:

1. Read arbitrary local files (e.g., )
2. Access internal services
3. Load attacker-controlled keys to bypass JWT signature verification

## Solution

Add a scheme whitelist check in  that only allows  and  URIs.

## Changes

- : Added  import and scheme validation
- : Added 5 new tests

## Testing

All 26 tests pass:

## Impact

- **Affected versions:** All versions &lt;= 2.12.1
- **CWE:** CWE-918 (Server-Side Request Forgery)

### DIFF
--- a/jwt/jwks_client.py
+++ b/jwt/jwks_client.py
@@ -6,6 +6,7 @@ from functools import lru_cache
 from ssl import SSLContext
 from typing import Any
 from urllib.error import HTTPError, URLError
+from urllib.parse import urlparse
 
 from .api_jwk import PyJWK, PyJWKSet
 from .api_jwt import decode_complete as decode_token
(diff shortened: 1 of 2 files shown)


jpadilla:
Addressed in 2.13.0

PR #1176, opened 2026-05-26
Add allowed_schemes to PyJWKClient

### LINKED ISSUE #1175: Define Allowed Schemes at PyJWKClient init
# Summary
PyJWKClient has the allowed schemes hard coded to "http" and "https". Developers have legitimate reasons to allow other schemes or not allow "http" in order to implement sub-classes for interfacing with other URI schemes such as an S3 bucket.
Code reference: https://github.com/jpadilla/pyjwt/blob/7144e4534c34810f4525dc4578a32addd8212cff/jwt/jwks_client.py#L78

# Observed Behavior
There is no argument to pass PyJWKClient that allows different URI schemes.
```python
from jwt import PyJWKClient
client = PyJWKClient("s3://my-bucket/jwks.json")
# [ERROR] PyJWKClientError: Invalid JWKS URI scheme 's3': only 'http' and 'https' are supported.
# Traceback...
```
This was not an error in PyJWT 2.12.0

# Proposed Solution
Add an optional `allowed_schemes` argument to PyJWKClient.
Fixed by #1176 

###  Environment
PyJWT 2.13.0
Python 3.12

davisjk:
# Summary
Fixes #1175 
PyJWKClient has the allowed schemes hard coded to "http" and "https". Developers have legitimate reasons to allow other schemes or not allow "http" in order to implement sub-classes for interfacing with other URI schemes such as an S3 bucket.

# Testing
- [x] This doesn't raise an error
```python
from jwt import PyJWKClient
client = PyJWKClient("s3://my-bucket/jwks.json", allowed_schemes=["s3"])
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,12 @@ This project adheres to `Semantic Versioning <https://semver.org/>`__.
 `Unreleased <https://github.com/jpadilla/pyjwt/compare/2.13.0...HEAD>`__
 ------------------------------------------------------------------------
 
+Fixed
+~~~~~
+
+- Callers can define allowed URI schemes in ``PyJWKClient``
+  by @davisjk in `#1176 <https://github.com/jpadilla/pyjwt/pull/1176>`__.
+
 `v2.13.0 <https://github.com/jpadilla/pyjwt/compare/2.12.1...2.13.0>`__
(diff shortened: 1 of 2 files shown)


davisjk:
@auvipy Thanks for the findings fix

PR #1180, opened 2026-06-03
Set hash_alg on OKPAlgorithm so compute_hash_digest works for EdDSA

### LINKED ISSUE #1097: EdDSA hash not working
`compute_hash_digest` does not work for `EdDSA`.

## Expected Result

I'd expect `compute_hash_digest` for `EdDSA` to work.

## Actual Result

It fails with `NotImplementedError`. This was originally reported at https://github.com/python-social-auth/social-core/issues/1306, so there might be something wrong in python-social-auth as well.

## Reproduction Steps

```console
(test) /tmp$ uv pip install 'pyjwt[crypto]'
Using Python 3.11.2 environment at: test
Resolved 4 packages in 313ms
░░░░░░░░░░░░░░░░░░░░ [0/4] Installing wheels...                                                                                                                                                                    
Installed 4 packages in 51ms
 + cffi==2.0.0
 + cryptography==46.0.1
 + pycparser==2.23
 + pyjwt==2.10.1
(test) /tmp$ python
Python 3.11.2 (main, Apr 28 2025, 14:11:48) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import jwt
>>> algo = jwt.get_algorithm_by_name('EdDSA')
>>> algo.compute_hash_digest(b'')
  File "/tmp/test/lib/python3.11/site-packages/jwt/algorithms.py", line 153, in compute_hash_digest
    raise NotImplementedError
NotImplementedError
```

## System Information

```
{
  "cryptography": {
    "version": "46.0.1"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.11.2"
  },
  "platform": {
    "release": "6.1.0-38-amd64",
    "system": "Linux"
  },
  "pyjwt": {
    "version": "2.10.1"
  }
}
```

arpitjain099:
Fixes #1097

## Problem

`Algorithm.compute_hash_digest()` raises `NotImplementedError` for `EdDSA`:

```python
>>> import jwt
>>> algo = jwt.get_algorithm_by_name('EdDSA')
>>> algo.compute_hash_digest(b'')
NotImplementedError
```

`compute_hash_digest` reads `self.hash_alg` and raises when it is `None`. Every other algorithm class (`HMACAlgorithm`, `RSAAlgorithm`, `ECAlgorithm`) sets `self.hash_alg` in `__init__`, but `OKPAlgorithm.__init__` was just `pass`, so the attribute was never set. This breaks downstream callers that compute OIDC `at_hash` / `c_hash` for EdDSA-signed tokens (see the linked python-social-auth report in the issue).

## Fix

Set `self.hash_alg` on `OKPAlgorithm`, mirroring how `RSAAlgorithm` / `ECAlgorithm` expose their hash via a `ClassVar`. EdDSA defaults to Ed25519, which hashes with SHA-512 internally per RFC 8037 (JOSE registration) and RFC 8032 (Ed25519 = SHA-512). Sign and verify are unchanged.

The `"EdDSA"` algorithm is registered as a single keyless instance (`"EdDSA": OKPAlgorithm()`), so the curve (Ed25519 vs Ed448) is not known at construction time. SHA-512 is therefore used as the default. Ed448 uses SHAKE256 rather than SHA-512; supporting that cleanly would require knowing the curve at the point the digest is computed, which is out of scope here. SHA-512 is correct for the common Ed25519 case.

## Tests

Added `test_okp_can_compute_digest`, which asserts `compute_hash_digest(b"foo")` for `EdDSA` returns the known SHA-512 digest of `"foo"` and is 64 bytes long. On the current code this test fails with `NotImplementedError`; with the fix it passes. The full test suite stays green (`pytest tests/`), and `ruff check` / `ruff format --check` / `mypy jwt/algorithms.py` are clean.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,14 @@ This project adheres to `Semantic Versioning <https://semver.org/>`__.
 `Unreleased <https://github.com/jpadilla/pyjwt/compare/2.13.0...HEAD>`__
 ------------------------------------------------------------------------
 
+Fixed
+~~~~~
+
+- Set ``hash_alg`` on ``OKPAlgorithm`` so ``compute_hash_digest`` works for
+  ``EdDSA`` instead of raising ``NotImplementedError``. EdDSA (Ed25519) hashes
+  with SHA-512 internally per RFC 8037 / RFC 8032. `#1097
+  <https://github.com/jpadilla/pyjwt/issues/1097>`__
+
 `v2.13.0 <https://github.com/jpadilla/pyjwt/compare/2.12.1...2.13.0>`__
 -----------------------------------------------------------------------
 
--- a/jwt/algorithms.py
+++ b/jwt/algorithms.py
@@ -857,6 +857,8 @@ if has_crypto:
         This class requires ``cryptography>=2.6`` to be installed.
         """
 
+        SHA512: ClassVar[type[hashes.HashAlgorithm]] = hashes.SHA512
+
         _crypto_key_types = cast(
             tuple[type[AllowedKeys], ...],
             get_args(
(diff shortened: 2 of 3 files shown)


jpadilla (changes requested):
[P2] Keep token-hash selection tied to verified key context

The fixed digest added here is not valid for every EdDSA configuration PyJWT supports. The [OpenID Connect Token Hash Algorithm Implementer's Guide, draft 00, section 2](https://openid.net/specs/openid-connect-token-hash-algorithms-1_0-00.html#section-2) explains the underlying constraint: for polymorphic `EdDSA`, the token-hash digest must be selected from the key that validated the ID Token rather than inferred from the `alg` value alone. Because this handler does not have that key context when it is constructed, choosing one unconditional default cannot preserve that invariant across the supported configurations.

Please either preserve `NotImplementedError` when the necessary context is unavailable, or make `compute_hash_digest` key/parameter-set aware. Add a control for each supported EdDSA parameter set, and update the documented OIDC flow if callers need to supply key context. I have a failing control for an existing supported configuration and can share the detailed reproduction privately.

PR #1181, opened 2026-06-03
Normalize base64 alphabet in base64url_decode to avoid Py3.15 FutureWarning

### LINKED ISSUE #1167: FutureWarning: invalid character '/' in URL-safe Base64 data will be discarded in future Python versions
Summary.
With Python 3.15 one of the tests emits:

```
E           FutureWarning: invalid character '/' in URL-safe Base64 data will be discarded in future Python versions
```

This can be suppressed by passing `-W ignore::FutureWarning` to pytest. But a proper fix would be better.

Originally reported in Fedora in https://src.fedoraproject.org/rpms/python-jwt/pull-request/15

arpitjain099:
Fixes #1167.

## Problem

Under Python 3.15 one test emits:

```
FutureWarning: invalid character '/' in URL-safe Base64 data will be discarded in future Python versions
```

`base64url_decode` in `jwt/utils.py` pads the input and hands it straight to `base64.urlsafe_b64decode`. Historically that decoder has quietly accepted standard-alphabet characters (`+` and `/`). Python 3.15 now warns on them, and CPython has announced it will eventually *discard* those characters, which would silently corrupt the decoded bytes for any caller that still passes standard-Base64 data.

The single input that trips the warning today is the `test_compressed_jwt` fixture. Its payload segment was produced with JavaScript `.toString('base64')` (standard alphabet) and carries a literal `/`, so it is not a valid RFC 7515 JWT segment (which must be base64url, using `_` and no `=` padding).

## Two possible directions

There are two reasonable ways to fix this, and I want to flag both:

1. **Harden the decoder** so valid URL-safe input never warns and standard-alphabet input is handled deterministically.
2. **Fix only the test fixture** so it carries proper base64url, leaving the decoder untouched.

I did both, but I deliberately chose *normalisation* rather than *rejection* for the decoder. Rejecting `+`/`/` would be the strictest reading of the function's contract, but it would be a behaviour change: pyjwt has accepted standard-alphabet input for a long time, and some real-world tokens from non-compliant issuers may rely on that. Normalising (`+`->`-`, `/`->`_`) before the decode keeps every currently-decodable token decodable, produces identical bytes, removes the FutureWarning, and stays correct once CPython starts discarding the characters. If you would rather reject standard-alphabet input outright (stricter, but a behaviour change), I am happy to switch to that.

## Changes

- `jwt/utils.py`: `base64url_decode` now translates the standard alphabet to the URL-safe alphabet (via a module-level `bytes.maketrans`) before padding and decoding. No `+`/`/` ever reaches `urlsafe_b64decode`.
- `tests/test_compressed_jwt.py`: the fixture is rebuilt on the URL-safe alphabet. Because the HMAC signature is computed over the literal `header.payload` signing input, changing the payload segment required recomputing the HS256 signature and the expected signature bytes; both are updated.
- `tests/test_utils.py`: adds coverage that standard- and URL-safe-alphabet inputs decode to the same bytes, and that decoding valid URL-safe (and normalised standard) input raises no `FutureWarning`.
- `CHANGELOG.rst`: entry under Unreleased / Fixed.

## Verification

- `pytest tests/ -q` passes (367 passed, 4 pre-existing `no_crypto_required` skips).
- `pytest tests/ -q -W error::FutureWarning` passes, proving the warning is gone for the suite.
- `ruff check`, `ruff format --check`, and `mypy` (strict) are clean.

I was on Python 3.14, which does not yet emit the 3.15 FutureWarning, so I could not reproduce the exact warning string locally. I confirmed the mechanism instead: the old code path passes `/` straight to `urlsafe_b64decode`, while the new path never does, and the new unit tests fail under `-W error::FutureWarning` on any version that emits it.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,17 @@ This project adheres to `Semantic Versioning <https://semver.org/>`__.
 `Unreleased <https://github.com/jpadilla/pyjwt/compare/2.13.0...HEAD>`__
 ------------------------------------------------------------------------
 
+Fixed
+~~~~~
+
+- Normalise the standard Base64 alphabet (``+`` / ``/``) to the URL-safe
+  alphabet in ``base64url_decode`` so valid URL-safe input no longer trips
+  the Python 3.15 ``FutureWarning: invalid character '/' in URL-safe Base64
+  data`` and standard-alphabet input keeps decoding deterministically once
+  CPython starts discarding those characters. Also corrects the
+  ``test_compressed_jwt`` fixture, which carried a standard-Base64 payload
+  segment instead of a URL-safe one (`#1167 <https://github.com/jpadilla/pyjwt/issues/1167>`__).
+
 `v2.13.0 <https://github.com/jpadilla/pyjwt/compare/2.12.1...2.13.0>`__
 -----------------------------------------------------------------------
 
--- a/jwt/utils.py
+++ b/jwt/utils.py
@@ -22,8 +22,18 @@ def force_bytes(value: Union[bytes, str]) -> bytes:
         raise TypeError("Expected a string value")
 
 
+# Translate the standard Base64 alphabet to the URL-safe alphabet so that
+# callers passing standard-alphabet data (``+`` / ``/``) are normalised before
+# the decode. From Python 3.15 ``base64.urlsafe_b64decode`` emits a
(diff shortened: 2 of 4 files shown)


jpadilla:
Superseded by e6f48401001609a8f99e71fcaf355fb895d508a8

PR #1186, opened 2026-06-20
fix: raise PyJWTError (not TypeError) for non-numeric exp/nbf/iat claims

binggao1230:
## Summary

Decoding a JWT whose `exp`, `nbf`, or `iat` claim is a non-numeric, non-string value (a list, dict, or `null`) leaks an uncaught `TypeError` instead of the intended `PyJWTError` subclass:

```python
import jwt
token = jwt.encode({"exp": [1]}, "secret", algorithm="HS256")
jwt.decode(token, "secret", algorithms=["HS256"])
# TypeError: int() argument must be a string, a bytes-like object or a real number, not 'list'
```

The same happens for `nbf`/`iat` and for `{...}` / `null` values. Because the payload is attacker-controlled, a crafted token crashes any application that only catches `jwt.PyJWTError` (or a subclass) around `decode()` — bypassing the library's exception contract. This is the same bug class as the maintainer-acknowledged #1177 (crash decoding a JWT with a `null` `kid`).

## Cause

`_validate_iat`, `_validate_nbf`, and `_validate_exp` in `jwt/api_jwt.py` each do `int(payload[claim])` inside `try / except ValueError`. `int()` raises `ValueError` only for non-numeric **strings** (e.g. `"not-an-int"`, which the existing tests cover) — but a list/dict/`None` makes it raise `TypeError`, which the handlers don't catch.

## Fix

Broaden each handler to `except (ValueError, TypeError):`. No message or logic change — these values now raise the existing `DecodeError` (`exp`/`nbf`) or `InvalidIssuedAtError` (`iat`), matching the documented "must be an integer" contract.

Added a parametrized regression test covering `exp`/`nbf`/`iat` × `list`/`dict`/`None` (9 cases). Full suite: 374 passed, 4 skipped (pre-existing, `cryptography`-fixture, unchanged by this PR). `ruff check` / `ruff format --check` clean.

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,13 @@ This project adheres to `Semantic Versioning <https://semver.org/>`__.
 `Unreleased <https://github.com/jpadilla/pyjwt/compare/2.13.0...HEAD>`__
 ------------------------------------------------------------------------
 
+Fixed
+~~~~~
+
+- Raise the documented ``PyJWTError`` subclass instead of leaking a
+  ``TypeError`` when the ``exp``, ``nbf``, or ``iat`` claim decodes to a
+  non-numeric, non-string value such as a list, dict, or ``null``.
+
 `v2.13.0 <https://github.com/jpadilla/pyjwt/compare/2.12.1...2.13.0>`__
 -----------------------------------------------------------------------
 
--- a/jwt/api_jwt.py
+++ b/jwt/api_jwt.py
@@ -473,7 +473,7 @@ class PyJWT:
     ) -> None:
         try:
             iat = int(payload["iat"])
-        except ValueError:
+        except (ValueError, TypeError):
             raise InvalidIssuedAtError(
                 "Issued At claim (iat) must be an integer."
             ) from None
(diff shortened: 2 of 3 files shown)


binggao1230:
Addressed the Copilot overflow comments in `bcb66fd`.

The previous follow-up only added `OverflowError` handling for `iat`; `nbf` and `exp` now catch it as well, so `int(float("inf"))` no longer leaks a raw `OverflowError` from `decode()`.

I also expanded the existing malformed numeric-claim regression to include `float("inf")` across `exp`, `nbf`, and `iat`.

Verification:
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_api_jwt.py::TestJWT::test_decode_raises_clean_error_if_claim_is_non_numeric_type -q`
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_api_jwt.py -q`
- `python -m compileall -q jwt tests/test_api_jwt.py`
- `git diff --check`

auvipy (approved):

jpadilla (approved):
Reviewed at bcb66fd60ec80593def31db31f0e4210958d0b4c. No actionable findings. Focused regression coverage, crypto/no-crypto suites, strict typing, documentation checks, and an independent security challenge passed.

PR #1188, opened 2026-07-04
Reject a bare string for the decode() algorithms argument (#1117)

### LINKED ISSUE #1117: Issue with type checking of  algorithm parameter of jwt.decode methods
Summary.

## Expected Result

As per PyJWT Documentation decode method expects algorithm as a sequence type or none
 jwt.decode(jwt_payload, "secret", leeway=10, algorithms=["HS256"])

## Actual Result
But if I write the code like this it is working fine
payload = jwt.decode(
                jwt=auth_token,
                key=FlaskProdConfig.SECRET_KEY,
                algorithms='HS256'
            )
## Issue details
``` The decode function annotation:

algorithms: Sequence[str] | None = None


is only a type hint — Python itself doesn’t check it at runtime.
So algorithms="HS256" is syntactically valid — no error occurs just because of typing.

Now, if you passed "HS256", this expression still works because:

"HS256" in "HS256"  # → True


so right now 
right approach 
>>> import jwt
>>> key = "secret"
>>> encoded = jwt.encode({"some": "payload"}, key, algorithm="HS256")
>>> print(encoded)
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzb21lIjoicGF5bG9hZCJ9.4twFt5NiznN84AWoo1d7KO1T_yoc0Z6XOpOVswacPZg
>>> jwt.decode(encoded, key, algorithms=["HS256"])
{'some': 'payload'}

will be work 


and wrong approach 


>>> import jwt
>>> key = "secret"
>>> encoded = jwt.encode({"some": "payload"}, key, algorithm="HS256")
>>> print(encoded)
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzb21lIjoicGF5bG9hZCJ9.4twFt5NiznN84AWoo1d7KO1T_yoc0Z6XOpOVswacPZg
>>> jwt.decode(encoded, key, algorithms="HS256")
{'some': 'payload'}

that will also work

```

This command is available on PyJWT v2.8.0 and greater. Otherwise,
please provide some basic information about your system.

apoorvdarshan:
Closes #1117.

## Problem

The `algorithms` argument to `decode()` is documented and typed as `Sequence[str]`, but a bare string is silently accepted:

```python
>>> jwt.decode(token, key, algorithms="HS256")   # "works"
```

It "works" only by accident: the allow-list check is `alg not in algorithms`, and for a string that's a **substring** test rather than membership. So a token's `alg` only has to be a *substring* of the passed string to pass the allow-list, which is a footgun for a value that is a security-critical control (the allow-list is what prevents algorithm-confusion / `alg=none` attacks).

```python
>>> "HS2" in "HS256"     # True  -> substring, not membership
```

## Fix

Raise a clear `TypeError` early in `decode_complete()` when `algorithms` is a `str`, telling the caller to pass a sequence. Proper sequences (`list`, `tuple`, `set`) are unaffected, and the "no algorithms passed" path is unchanged.

```python
>>> jwt.decode(token, key, algorithms="HS256")
TypeError: The "algorithms" argument must be a sequence of strings (e.g. ["HS256"]), ...
>>> jwt.decode(token, key, algorithms=["HS256"])   # still works
```

## Testing

- Added `TestJWS::test_decode_rejects_str_algorithms` — a bare string raises `TypeError`; a list still decodes. It does *not* raise on `main` (`DID NOT RAISE`) and passes with this change.
- Full suite: `366 passed, 4 skipped` (the skips are the pre-existing crypto-optional tests); no existing test relied on passing a bare string. `flake8` and `mypy` are clean on the changed source.
- Added a `CHANGELOG.rst` entry under *Unreleased → Changed*.

## Note

This is a small **breaking** change: code that currently passes a bare string will now raise instead of silently working. Given `algorithms` is a security control, I went with failing loud, but if you'd prefer a softer path (a `DeprecationWarning` now, coercing the string to a single-element list, and the hard error in pyjwt 3), I'm happy to adjust.

---

*Disclosure: this change was prepared with the assistance of an AI tool (Claude Code). I reproduced the issue, implemented and verified the fix and test, ran the suite and linters, and take responsibility for the contribution and will respond to review feedback personally.*

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,14 @@ This project adheres to `Semantic Versioning <https://semver.org/>`__.
 `Unreleased <https://github.com/jpadilla/pyjwt/compare/2.13.0...HEAD>`__
 ------------------------------------------------------------------------
 
+Changed
+~~~~~~~
+
+- ``decode()`` now raises ``TypeError`` when the ``algorithms`` argument is a
+  bare string instead of a sequence of strings. A string was matched one
+  character at a time, so a token's algorithm only had to be a substring of it,
+  silently weakening the algorithm allow-list.
+
 `v2.13.0 <https://github.com/jpadilla/pyjwt/compare/2.12.1...2.13.0>`__
 -----------------------------------------------------------------------
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -9,7 +9,7 @@ Encoding & Decoding Tokens with HS256
     >>> import jwt
     >>> key = "secret"
     >>> encoded = jwt.encode({"some": "payload"}, key, algorithm="HS256")
-    >>> jwt.decode(encoded, key, algorithms="HS256")
+    >>> jwt.decode(encoded, key, algorithms=["HS256"])
     {'some': 'payload'}
 
 Encoding & Decoding Tokens with RS256 (RSA)
--- a/jwt/api_jws.py
+++ b/jwt/api_jws.py
@@ -233,6 +233,13 @@ class PyJWS:
                 RemovedInPyjwt3Warning,
                 stacklevel=2,
             )
+        if isinstance(algorithms, str):
+            raise TypeError(
+                'The "algorithms" argument must be a sequence of strings '
+                '(e.g. ["HS256"]), not a bare string. A string is matched one '
+                "character at a time, so a token's algorithm only has to be a "
+                "substring of it, which silently weakens the algorithm allow-list."
+            )
         merged_options: SigOptions
         if options is None:
             merged_options = self.options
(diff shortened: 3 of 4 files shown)


apoorvdarshan:
Updated the usage doctest to pass algorithms as a list, matching the new validation. Local doctests pass, and tests/test_api_jws.py passes (89 passed, 1 skipped).

godhunter98:
Hit this independently while pinning algorithms for OIDC ID token validation.

One more illustration of the substring semantics, since "HS2" in "HS256" uses a value nobody would pass: the empty string is a substring of every string, so with algorithms="HS256" a header of {"alg": ""} clears the alg not in algorithms check. It fails immediately afterwards at algorithm lookup, so there's no bypass , but it shows the allow-list isn't performing membership at all, which is worth a test case alongside the existing ones.

Worth noting for anyone skimming: the reverse is not true. algorithms="" rejects everything except alg=""; it doesn't accept everything.

The TypeError in this PR seems like the right shape, it fixes the footgun without narrowing Sequence[str] and breaking callers who pass tuples.

PR #1199, opened 2026-08-19
Support Ed25519 and Ed448 fully specified algorithms (RFC 9864)

### LINKED ISSUE #1190: Support RFC 9864 Fully specified algorithms (Ed25519, Ed448, etc.)
## Feature request

Add support for fully specified algorithms such as `Ed25519`.

These algorithms were added by [RFC 9864](https://datatracker.ietf.org/doc/html/rfc9864).

You can see the full list of registered algorithms here: https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms

## My issue

I tried using `Ed25519` with the `PyJWKClient` and it didn't work. The key was [silently skipped](https://github.com/jpadilla/pyjwt/blob/7144e4534c34810f4525dc4578a32addd8212cff/jwt/api_jwk.py#L151), and then I got a confusing `keyset has no key for kid` error.

The actual error was:

```
jwt.exceptions.PyJWKError: Unable to find an algorithm for key: {'kty': 'OKP', 'crv': 'Ed25519', 'alg': 'Ed25519', 'use': 'sig', 'x': '--OMITTED--', 'kid': '--OMITTED--'}
```

Then I tried to workaround the issue by manually re-registering `EdDSA` as `Ed25519`, but that didn't work because `PyJWK` [only consults the default algorithms](https://github.com/jpadilla/pyjwt/blob/7144e4534c34810f4525dc4578a32addd8212cff/jwt/api_jwk.py#L76C30-L76C52).

If it was possible to override the algorithms used, that would fix my issue too.

wbolster:
This adds support for ‘Ed25519’ and ‘Ed448’ fully-specified algorithms
as specified in RFC 9864 <https://datatracker.ietf.org/doc/html/rfc9864>,
‘Fully-Specified Algorithms for JSON Object Signing and
Encryption (JOSE) and CBOR Object Signing and Encryption (COSE)’.

Technically these are the same as the already supported ‘EdDSA’
algorithm, but with the used curve made explicit. When a fully-specified
algorithm value is used, the actual key is validated against the
expected curve type.

The original polymorphic ‘EdDSA’ identifier (which doesn't specify the
exact curve but is otherwise the same) keeps working as before.

Updated docs accordingly, and organised the overview of supported
algorithms per ‘family’ to make it easier to read.

Implementation notes:

- OKPAlgorithm now takes an optional ‘expected_curve’ argument, similar
  to ECAlgorithm.

- OKPAlgorithm.prepare_key() now checks the expected curve. This method
  now uses a single code path for validation, and curve validation is
  just a few if/else statements, without a separate helper (cf.
  ECAlgorithm._validate_curve() for EC keys which has a more complex
  control flow).

- The PyJWK constructor only handled ‘crv: Ed25519’ but not ‘crv:
  Ed448’. This seems to be an oversight from when Ed448 was added,
  because before that only Ed25519 was supported. Now it handles both.

- Example EdDSA keys for testing now include the ‘alg’ parameter, as
  recommended by RFC 9864 §7 (Security Considerations). Tests using
  these fixtures cover variants such as correct, incorrect, and missing
  ‘alg’ values.

Closes #1190.

### DIFF
--- a/jwt/api_jwk.py
+++ b/jwt/api_jwk.py
@@ -37,9 +37,9 @@ class PyJWK:
         if not algorithm and isinstance(self._jwk_data, dict):
             algorithm = self._jwk_data.get("alg", None)
 
+        crv = self._jwk_data.get("crv", None)
         if not algorithm:
             # Determine alg with kty (and crv).
-            crv = self._jwk_data.get("crv", None)
             if kty == "EC":
                 if crv == "P-256" or not crv:
                     algorithm = "ES256"
@@ -58,7 +58,7 @@ class PyJWK:
             elif kty == "OKP":
                 if not crv:
                     raise InvalidKeyError(f"crv is not found: {self._jwk_data}")
-                if crv == "Ed25519":
+                if crv in ("Ed25519", "Ed448"):
                     algorithm = "EdDSA"
                 else:
                     raise InvalidKeyError(f"Unsupported crv: {crv}")
@@ -69,6 +69,14 @@ class PyJWK:
             raise MissingCryptographyError(
                 f"{algorithm} requires 'cryptography' to be installed."
             )
+
+        # For the OKP key type, the optional ‘alg’ can be either a polymorphic
+        # algorithm identifier (EdDSA), or a fully-specified algorithm
+        # identifier (Ed25519 or Ed448), in which case it must match the value
+        # in the required curve field. (This check cannot be done above because
+        # the OKP inference branch above only runs when the ‘alg’ is missing.)
+        if kty == "OKP" and algorithm in ("Ed25519", "Ed448") and algorithm != crv:
+            raise InvalidKeyError(f"Unsupported crv for {algorithm}: {crv}")
 
         self.algorithm_name = algorithm
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,12 @@ This project adheres to `Semantic Versioning <https://semver.org/>`__.
 `Unreleased <https://github.com/jpadilla/pyjwt/compare/2.13.0...HEAD>`__
 ------------------------------------------------------------------------
 
+Added
+~~~~~
+
+- Support fully-specified algorithm identifiers ``Ed25519`` and
+  ``Ed448`` as defined in RFC 9864.
+
 `v2.13.0 <https://github.com/jpadilla/pyjwt/compare/2.12.1...2.13.0>`__
 -----------------------------------------------------------------------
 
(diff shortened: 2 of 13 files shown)


wbolster:
note that i've included 2 preparatory commits. if maintainers prefer, i can spin those into separate pull requests. just let me know!

mangrisano:
LGTM.

jwt/api_jwk.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, jwk_data: JWKDict, algorithm: str | None = None) -> None:
             elif kty == "OKP":
                 if not crv:
                     raise InvalidKeyError(f"crv is not found: {self._jwk_data}")
-                if crv == "Ed25519":
+                if crv in ("Ed25519", "Ed448"):
```
jpadilla, line 61: [P2] Support fully specified tokens with algorithm-less OKP JWKs

When `alg` is omitted, this branch binds every OKP key to `EdDSA`. The later exact PyJWK binding then rejects a valid `Ed25519` or `Ed448` token before signature verification, even though RFC 9864 section 5 permits the JWK `alg` member to be omitted. I reproduced this for both curves through `PyJWK`; the same key-selection path is used by `PyJWKClient`. Please distinguish an explicit JWK algorithm constraint from an inferred default so an algorithm-less key can verify the caller-allowed, curve-compatible fully specified identifier while retaining legacy `EdDSA` support. Add PyJWK/PyJWKClient controls for both curves and both legacy and fully specified identifiers.

jpadilla (changes requested):
Security classification: `security-boundary-affected`. The algorithm-choice, OKP key-material, and asymmetric-JWK paths were reviewed. One compatibility issue needs correction before approval.

jwt/api_jwk.py
jpadilla: Minimal repro for both `Ed25519` and `Ed448`:

1. Load a matching OKP public JWK without its optional `alg` member.
2. Construct it with `PyJWK.from_dict()`.
3. Verify a valid token whose header uses the curve's fully specified algorithm.

Expected: verification succeeds because the curve matches and the caller explicitly allowed that algorithm.

Actual:

```text
Ed25519 bound-as EdDSA
InvalidAlgorithmError: Token algorithm 'Ed25519' does not match the key's algorithm 'EdDSA'

Ed448 bound-as EdDSA
InvalidAlgorithmError: Token algorithm 'Ed448' does not match the key's algorithm 'EdDSA'
```

Matching controls succeed when the JWK includes `alg`, when `PyJWK` receives an explicit algorithm override, or when the token uses legacy `EdDSA`. [RFC 9864 section 5](https://www.rfc-editor.org/rfc/rfc9864.html#section-5) permits the JWK `alg` member to be omitted.

PR #1201, opened 2026-08-24
Catch http.client.HTTPException in PyJWKClient.fetch_data

### LINKED ISSUE #1013: `JWKClient.get_jwk_set()` raises `http.client.IncompleteRead`
When trying to integrate into an internal JWKS endpoint, `PyJWKClient.get_jwk_set()` returns `http.client.IncompleteRead`.

I can fetch the content via curl.

As an aside, I also successfully get the response using both `htpx.get()` and `requests.get()`. I can convert that result into a `jwt.PyJWKSet` (using `jwt.PyJWKSet.from_dict(response.json())`. Then the rest of the validation works. 

Additionally, I'm able to get the same results just using `urllib.request.Request` (as done [here](https://github.com/jpadilla/pyjwt/blob/master/jwt/jwks_client.py#L55-L59)) directly. 

## Expected Result

The call should return the key set. I've verified. Ideally, the JWKClient would handle retrying this within the `PyJWKClient.fetch_data` method. 

<details>
<summary>This example works after one retry. (expand)</summary>

```
import urllib.request
import http.client

def fetch_url(url, retries=3):
    for attempt in range(retries):
        try:
            with urllib.request.urlopen(url) as response:
                return response.read()
        except http.client.IncompleteRead as e:
            if attempt < retries - 1:
                print(f"Incomplete read, retrying... ({attempt + 1}/{retries})")
                continue
            else:
                raise
    raise RuntimeError("Max retries exceeded")

try:
    content = fetch_url(url)
    print("Content fetched successfully")
except Exception as e:
    print(f"Failed to fetch content: {e}")
```
</details>

## Actual Result

```
  File "/home/scotchneat/dev/temp/jwt_sandbox/.venv/lib/python3.12/site-packages/jwt/jwks_client.py", line 74, in get_jwk_set
    data = self.fetch_data()
           ^^^^^^^^^^^^^^^^^
  File "/home/scotchneat/dev/temp/jwt_sandbox/.venv/lib/python3.12/site-packages/jwt/jwks_client.py", line 57, in fetch_data
    jwk_set = json.load(response)
              ^^^^^^^^^^^^^^^^^^^
  File "/home/scotchneat/.pyenv/versions/3.12.2/lib/python3.12/json/__init__.py", line 293, in load
    return loads(fp.read(),
                 ^^^^^^^^^
  File "/home/scotchneat/.pyenv/versions/3.12.2/lib/python3.12/http/client.py", line 495, in read
    s = self._safe_read(self.length)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/scotchneat/.pyenv/versions/3.12.2/lib/python3.12/http/client.py", line 642, in _safe_read
    raise IncompleteRead(data, amt-len(data))
http.client.IncompleteRead: IncompleteRead(14594 bytes read, 5172 more expected)
```

## Reproduction Steps

```python
import jwt
import os

url = os.getenv('JWKS_URL')

cli = jwt.PyJWKClient(url)
key_set = get_jwk_set()
```

## System Information

    $ python -m jwt.help

```
{
  "cryptography": {
    "version": "43.0.3"
  },
  "implementation": {
    "name": "CPython",
    "version": "3.12.2"
  },
  "platform": {
    "release": "6.8.0-47-generic",
    "system": "Linux"
  },
  "pyjwt": {
    "version": "2.9.0"
  }
}
```

aroh3006:
Fixes #1013.

`fetch_data()` only catches `URLError` and `TimeoutError`, but a connection that drops mid-response makes `json.load(response)` raise `http.client.IncompleteRead`, which is neither. That escapes as a raw stdlib exception instead of the `PyJWKClientConnectionError` the method's own docstring promises for "the HTTP request fails".

Widened the except clause to also catch `http.client.HTTPException`, same wrapping as the existing cases.

Added a regression test mocking a response whose `.read()` raises `IncompleteRead`; fails on main, passes with the fix. Ran the full test suite, ruff, and mypy on the changed files, all clean.


CHANGELOG.rst
```diff
@@ -87,6 +87,14 @@ Fixed
   ``TypeError`` when the ``exp``, ``nbf``, or ``iat`` claim decodes to a
   non-numeric, non-string value such as a list, dict, or ``null``.
 
+Fixed
+~~~~~
+
+- Wrap ``http.client.HTTPException`` (e.g. ``IncompleteRead`` from a
```
jpadilla: [P2] This note is beneath the v2.14.0 heading, so it attributes the fix to an already released version and leaves it out of the next release notes. Please move it into the existing Unreleased / Fixed section.

jpadilla (commented):
One release-note placement issue to address below.

CHANGELOG.rst
aroh3006: Moved it into the Unreleased / Fixed section.

jpadilla (approved):

PR #1202, opened 2026-09-02
Add support for Python 3.15

kytta:
Currently blocked by #1167


.github/workflows/main.yml
```diff
@@ -29,6 +29,7 @@ jobs:
           - "3.12"
           - "3.13"
           - "3.14"
+          - "3.15"
```
auvipy, line 32: isnt it still in dev or rc version?

kytta: It is! It's in RC tho, meaning that the API and ABI are stable, so it's safe to test under it and declare support for it. I usually write that in PR body, but didn't do that here because it's not ready yet :)

jpadilla:
@kytta can you update from main?

kytta:
Rebased 👍🏻  

Tests now pass locally (?); I'm not sure what fixed those given that #1167 is still open ¯\\_(ツ)_/¯

jpadilla (approved):

PR #1204, opened 2026-09-07
Fix `get_jwk_set` for `PyJWKSet` values in the JWK Set cache

### LINKED ISSUE #914: The JWKS endpoint did not return a JSON object when keys are found in the cache

mmdverse:
Fixes #914.

`JWKSetCache.put()` documents `PyJWKSet` as the cached value, so callers can pre-populate the cache (e.g. with a pre-fetched JWKS to avoid a network round-trip). `get_jwk_set()` only accepted a `dict` from the cache, so a cache hit holding a `PyJWKSet` raised:

```
PyJWKClientError: The JWKS endpoint did not return a JSON object
```

which is misleading, since no endpoint call is involved.

This returns cached `PyJWKSet` values directly, before the `dict` check:

- the internal `fetch_data()` flow (which stores the raw dict) is unchanged
- other invalid cache values still raise `PyJWKClientError` as before
- adds a regression test (fails before the fix, passes after)

### DIFF
--- a/jwt/jwks_client.py
+++ b/jwt/jwks_client.py
@@ -195,6 +195,10 @@ class PyJWKClient:
         if data is None:
             data = self.fetch_data()
 
+        # The cache may hold a PyJWKSet (its documented value type, see
+        # JWKSetCache.put) as well as the raw dict stored by fetch_data().
+        if isinstance(data, PyJWKSet):
+            return data
         if not isinstance(data, dict):
             raise PyJWKClientError("The JWKS endpoint did not return a JSON object")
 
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
please fix the failing unit tests

jpadilla:
addressed in #1208

